PR #9, opened 2024-10-21
Implement PlaceholderEmpty component

spli02:


nighto (approved):
LGTM, but a suggestion to improve: it'd be nice if the documentation / stories state the default values for the element, such as size large, icon graph etc.

src/components/graphics/PlaceholderEmpty/PlaceholderEmpty.tsx
```diff
@@ -0,0 +1,104 @@
...
+    customIcon = null,
+    title = '',
+    subtitle = '',
+    actions = null,
+    children,
+    ...propsRest
+  } = props;
+  
+  const icon = ((): PlaceholderEmptyIconType => {
+    switch (iconType) {
```
imadbakir, line 59: shouldn't the component be re-usable with any available icon?

imadbakir (changes requested):

src/components/graphics/PlaceholderEmpty/PlaceholderEmpty.tsx
spli02: In Figma, it looks like these 4 icons will be built-in images, so I defined `PlaceholderEmptyIconType` to make this clear, similar to the icons in `Alert.tsx`. Users can also customize it through the `customIcon` prop. Please let me know if you have any better suggestions.

imadbakir: IMHO, the icons that are present in the Figma are just example use cases, I don't think we need some sort of pre-defined set of default icons to choose from. 
I think we just need an icon of type IconName, 
and an option to pass a custom icon (in case consumer want to use some other type of graphic, png for example etc..)

PR #24, opened 2024-11-17
[Switch] Add transitions for Switch thumb and track colors

mkrause:
When I tried to get color transitions to work back in Chrome v125, it seemed to not work with light-dark(). I tried it again now, and the colors correctly transition. Firefox and Safari also work correctly. So either this got fixed in the meantime, or maybe I did something wrong. Maybe the issue was `initial-value` should have been a plain color expression rather than `light-dark()`, since according to the spec `initial-value` must be "computationally independent", and `light-dark()` is not.

In any case, this PR fixes it so that now Switch has color transitions.

### DIFF
--- a/src/components/forms/controls/Switch/Switch.module.scss
+++ b/src/components/forms/controls/Switch/Switch.module.scss
@@ -1,10 +1,10 @@
 
 @use '../../../../styling/defs.scss' as bk;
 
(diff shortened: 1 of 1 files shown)


nighto (approved):
LGTM :)

PR #28, opened 2024-11-19
Tag component + InputField story using it

nighto:


src/components/forms/controls/Tag/Tag.module.scss
```diff
@@ -0,0 +1,27 @@
...
+    padding: 2px 0 3px bk.$spacing-2;
+
+    .bk-tag__icon {
+      width: 7px;
+      height: 7px;
+      color: bk.$theme-tag-icon-default;
+      cursor: pointer;
+      padding: 2px bk.$spacing-2 2px bk.$spacing-2;
+    }
+  }
```
mkrause, line 26: There's a bunch of `px` values above that should ideally be variables

mkrause (changes requested):

src/components/forms/fields/InputField/InputField.stories.tsx
```diff
@@ -55,3 +55,34 @@ export const InvalidInput: Story = {
...
+      }
+    };
+
+    return (
+      <InputField
+        tags={tags}
+        value={inputText}
+        onKeyUp={onKeyUp}
+        onChange={onChange}
+        placeholder={''}
```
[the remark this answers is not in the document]
nighto: Yeah my idea was to make it easier to test, have it prefilled with some text, so you can press Enter and see it becoming a tag. I'll change it to `""`.

src/components/forms/controls/Tag/Tag.tsx
```diff
@@ -0,0 +1,46 @@
...
+export type TagProps = ComponentProps<'div'> & {
+  /** Whether this component should be unstyled. */
+  unstyled?: undefined | boolean,
+
+  /** The text displayed inside the tag. */
+  value: string,
+};
+
+/**
+ * A tag component, meant to be used within Input fields.
```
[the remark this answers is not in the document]
nighto: Fair enough, it was just that the example on Figma was present on Input, replacing it.

src/components/forms/controls/Tag/Tag.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+
+export { cl as TagClassNames };
+
+export type TagProps = ComponentProps<'div'> & {
+  /** Whether this component should be unstyled. */
+  unstyled?: undefined | boolean,
+
+  /** The text displayed inside the tag. */
+  value: string,
```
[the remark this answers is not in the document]
nighto: All examples were with simple text, but indeed it just reuses the text inside as `{value}`, so I'll change it to JSX.

mkrause (changes requested):

src/components/forms/fields/InputField/InputField.tsx
```diff
@@ -30,8 +30,11 @@ export type InputFieldProps = ComponentProps<'input'> & {
   /** Tags to be displayed inside the input field */
   tags?: undefined | string[],
 
-  /** Callback to remove a specific Tag, passed down to Tag component. */
-  tagRemoveCallback?: (index: number) => void,
+  /** Callback to update the input value. Internally hooks to onChange */
+  onUpdate?: undefined | ((arg0: string) => void),
+
+  /** Callback to update the tags. Internally hooks to onKeyUp */
+  onUpdateTags?: undefined | ((arg0: string[]) => void),
```
mkrause, line 37: Instead of `arg0` please give these some meaningful names. For example `value` and `tags`, respectively.

src/components/forms/fields/InputField/InputField.tsx
```diff
@@ -43,7 +46,8 @@ export const InputField = (props: InputFieldProps) => {
     labelProps = {},
     wrapperProps = {},
     tags = [],
-    tagRemoveCallback,
+    onUpdate = null,
+    onUpdateTags = null,
```
mkrause, line 50: Should probably be `undefined` instead of `null` (i.e. just remove the `= null` defaults), since the types for these props use `undefined`.

src/components/forms/fields/InputField/InputField.tsx
```diff
@@ -55,6 +59,41 @@ export const InputField = (props: InputFieldProps) => {
...
+    if (inputProps.onKeyUp) {
+      inputProps.onKeyUp(e);
+    }
+    // then return value to onUpdateTags
+    if (onUpdateTags && onUpdate) {
+      const { value } = inputProps;
+      if (e.key === 'Backspace' && value === '') {
+        onUpdateTags(tags.slice(0,-1));
+      }
+      if (e.key === 'Enter' && value !== '') {
```
mkrause, line 84: What should we do with blankspace, e.g. `value === '   '`? Should we maybe do a `value.trim()` above?

src/components/forms/fields/InputField/InputField.tsx
```diff
@@ -55,6 +59,41 @@ export const InputField = (props: InputFieldProps) => {
...
+        onUpdateTags([...tags as string[], value as string]);
```
mkrause, line 85: Are the casts necessary here? They seem unsafe. If it's to get rid of an `undefined` case can do something like `[...(tags || [])]`. I'm looking at this on GitHub so can't see the inferred types at this point.

mkrause (changes requested):

src/components/forms/fields/InputField/InputField.tsx
nighto: good point

mkrause (approved):

Issue #29, opened 2024-11-19
Fix type errors in Tabs.stories.tsx


mkrause:
There are a few type errors in `src/components/navigations/Tabs/Tabs.stories.tsx`. Currently these are ignored in `tsconfig.json` in order to unblock CI. Need to resolve these errors, and then remove the `exclude`.

PR #30, opened 2024-11-19
Fix all TypeScript errors (except Tabs which is ignored for now)

mkrause:
I've resolved all type errors, except the ones in Tabs.stories.tsx which needs a bit of investigation. Ignored these for now and created an issue: https://github.com/fortanix/baklava/issues/29

This will allow us to start enforcing zero type errors in CI.


nighto (approved):
LGTM, no warnings running `npm run check-types`.

PR #42, opened 2024-12-05
Datepicker

nighto:


nighto:
TODO: Remove whitespace from package files.

nighto:
Next steps:

- [ ] Remove errors also from DateTimeRangePicker
- [ ] Refactor TimePicker to work with string instead of date object, exposing an object with `hours` and `minutes`
- [ ] Create an example with all together.

mkrause:
- [ ] Need to make all the stories use a hardcoded date/time so that it doesn't cause issues with visual testing

mkrause:
There's a warning in the console, let's fix this. Basically, need to spell out the class in full rather than relying on Sass-style concatenation:

```
warnings when minifying css:
▲ [WARNING] Cannot use type selector "--selected" directly after nesting selector "&" [css-syntax-error]

    <stdin>:2441:28:
      2441 │ .react-datepicker__day:not(&--selected):hover {
           │                             ~~~~~~~~~~
           ╵                             :is(--selected)

  CSS nesting syntax does not allow the "&" selector to come before a type selector. You can wrap this selector in ":is(...)" as a workaround. This restriction exists to avoid problems with SASS nesting, where the same syntax means something very different that has no equivalent in real CSS (appending a suffix to the parent selector).
```

src/components/forms/controls/DatePicker/DatePicker.module.scss
```diff
@@ -0,0 +1,213 @@
...
+  border-radius: 0 bk.$border-radius-circle bk.$border-radius-circle 0;
+}
+
+:global(.react-datepicker__day--range-start.react-datepicker__day--range-end),
+:global(.react-datepicker__day--range-start.react-datepicker__day--range-end:not([aria-disabled="true"]):hover) {
+  border-radius: bk.$border-radius-circle;
+}
+
+/* stylelint-disable-next-line */
+:global(.react-datepicker__day:not(&--selected):hover) {
```
mkrause, line 159: We shouldn't be using Sass-style string concatenation here (`&--selected`), with native CSS nesting this is no longer supported. Can instead write `.react-datepicker__day--selected` in full.

src/components/forms/controls/DateTimePicker/DateTimePicker.tsx
```diff
@@ -0,0 +1,69 @@
...
+      const [newHours, newMinutes] = time.split(':');
+      if (newHours) {
+        newDate.setHours(Number(newHours));
+      }
+      if (newMinutes) {
+        newDate.setMinutes(Number(newMinutes));
+      }
+      onChange(newDate);
+    }
+  }
```
mkrause, line 48: Add missing semicolon

src/components/forms/controls/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export type TimePickerProps = ComponentProps<'input'> & {
+  /** Whether this component should be unstyled. */
+  unstyled?: undefined | boolean,
+
+  /** An optional class name to be appended to the class list. */
+  className?: ClassNameArgument,
+  
+  /** A time string as defined to be used by input type="time", with the hh:mm format. */
+  time: string,
```
mkrause, line 22: Can we instead have `time` be an object with hours and minutes?

src/components/forms/controls/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,67 @@
...
+        customInput={
+          <Input className={cx(
+            { [cl['bk-datepicker--input']]: !unstyled },
+          )}/>
+        }
+        onCalendarClose={() => setIsOpen(false)}
+        onCalendarOpen={() => setIsOpen(true)}
+        {...propsRest}
+      />
+      <Icon icon={`caret-${isOpen ? 'up' : 'down'}`} className={cx(cl['bk-datepicker--caret'])}/>
```
mkrause, line 64: The caret here is current not clickable. If you click this I would expect it to trigger the picker.

src/components/forms/controls/DatePicker/DatePicker.module.scss
```diff
@@ -0,0 +1,207 @@
+/* Copyright (c) Fortanix, Inc.
+|* This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy of
+|* the MPL was not distributed with this file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+@use '../../../../styling/defs.scss' as bk;
+
+// TODO: copied manually from old Baklava, where should we put this?
+$z-index-date-picker: 100;
```
mkrause, line 8: Ideally we'd use popover for this. But not sure if we can make that work with this library.

If we start having multiple z-index values again I think we should have a single file for it like we did in previous Baklava.

src/components/forms/controls/DatePicker/DatePicker.module.scss
```diff
@@ -0,0 +1,207 @@
...
+
+:global(.react-datepicker__triangle),
+:global(.react-datepicker__triangle::before) {
+  display: none;
+}
+
+:global(.react-datepicker__header) {
+  background: transparent;
+  border: 0;
+  color: #081353; // TODO hardcoded color
```
mkrause, line 36: Is this even used? Doesn't seem like this color has any effect. Should we remove it?

src/components/forms/controls/DatePicker/DatePicker.stories.tsx
```diff
@@ -0,0 +1,45 @@
...
+    Story => <form onSubmit={event => { event.preventDefault(); }}><Story/></form>,
+  ],
+} satisfies Meta<DatePickerArgs>;
+
+export const Standard: Story = {
+  render: (args) => {
+    const [startDate, setStartDate] = React.useState<Date | null>(new Date(2024, 11, 12, 9, 41));
+
+    return (
+      // with a fixed height to avoid hiding the bottom part of the calendar
```
mkrause, line 34: One of the reasons a popover would be great here instead of z-index 🙂

src/components/forms/controls/DatePicker/DatePicker.stories.tsx
```diff
@@ -0,0 +1,45 @@
...
+      <div style={{height: '500px'}}>
+        <DatePicker
+          {...args}
+          selected={startDate}
+          onChange={(date: Date | null) => setStartDate(date)}
```
mkrause, line 39: Can you wrap the function body here in a block?

src/components/forms/controls/DatePicker/DatePickerRange.stories.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+    return (
+      // with a fixed height to avoid hiding the bottom part of the calendar
+      <div style={{height: '500px'}}>
+        <DatePickerRange
+          {...args}
+          selected={startDate}
+          onChange={onChange}
+          startDate={startDate as Date}
+          endDate={endDate as Date}
```
mkrause, line 47: This seems unsafe. Why are `startDate` and `endDate` nullable in the first place?

src/components/forms/controls/DatePicker/DatePickerRange.stories.tsx
```diff
@@ -0,0 +1,58 @@
...
+          inline={true}
```
mkrause, line 48: What does `inline` do here?

src/components/forms/controls/DatePicker/DatePickerRange.tsx
```diff
@@ -0,0 +1,57 @@
...
+  return (
+    <div className={cx(
+      'bk',
+      { [cl['bk-datepicker']]: !unstyled },
+      className,
+    )}>
+      <ReactDatePicker
+        selectsRange
+        // everything else is the same
+        dateFormat="MM/dd/yyyy"
```
mkrause, line 50: This should be a prop, so consumers can customize the format. (Also for `DatePicker`)

src/components/forms/controls/DateTimePicker/DateTimePicker.module.scss
```diff
@@ -0,0 +1,20 @@
+/* Copyright (c) Fortanix, Inc.
+|* This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy of
+|* the MPL was not distributed with this file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+@use '../../../../styling/defs.scss' as bk;
+
+@layer baklava.components {
+  .bk-datetimepicker {
```
mkrause, line 8: Can you rename this to `date-time-picker`? Also for `datepicker` and `timepicker`

src/components/forms/controls/DateTimePicker/DateTimePicker.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+type GenericProps = {
+  /** Whether this component should be unstyled. */
+  unstyled?: undefined | boolean,
+
+  /** An optional class name to be appended to the class list. */
+  className?: undefined | ClassNameArgument,
+  
+  /** A Date object to hold the date and time. */
+  date: Date | null,
```
mkrause, line 23: I think we can make this a required prop. The consumer can always initialize it to a default value like the current date/time.

Also for the other components.

src/components/forms/controls/TimePicker/TimePicker.module.scss
```diff
@@ -0,0 +1,18 @@
...
+
+@layer baklava.components {
+  .bk-timepicker {
+    @include bk.component-base(bk-timepicker);
+    line-height: 1.8;
+  }
+
+  .bk-timepicker--input {
+    color: bk.$theme-form-text-default;
+    // a manual size increase to accomodate the clock icon without clipping the text
```
mkrause, line 15: accomodate → accommodate

Also if you can make the comments sentence case for consistency

src/components/forms/controls/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+import { Input } from '../Input/Input.tsx';
+
+import cl from './TimePicker.module.scss';
+
+
+export type Time = {
+  hours: number,
+  minutes: number,
+}
```
mkrause, line 17: Add a semicolon here

mkrause (changes requested):

src/components/forms/controls/DatePicker/DatePickerRange.stories.tsx
nighto: `inline` makes it open directly as a datepicker, instead of being an input element that when clicked reveal the datepicker.

src/components/forms/controls/DateTimePicker/DateTimePicker.tsx
nighto: The problem is that the library expects it to be `Date | null`, I tried setting it as a plain `Date` but then we start to get type mismatches. Maybe setting it as `Date` on our code and then passing it down to the library with `as Date | null`?

src/components/forms/controls/DatePicker/DatePicker.module.scss
nighto: Moved it to variables.scss.

src/components/forms/controls/DatePicker/DatePickerRange.stories.tsx
nighto: Just to document here as well: there is a bug in the original library, already fixed and merged to master but not released yet. https://github.com/Hacker0x01/react-datepicker/pull/5260

src/components/forms/controls/DateTimePicker/DateTimePicker.tsx
nighto: Documenting here our conversation: `Date | null` is a valid value for uninitialized calendar.

src/styling/lib/react-datepicker.scss
```diff
@@ -0,0 +1,178 @@
...
+  font-weight: bk.$font-weight-semibold;
+  text-transform: uppercase;
+}
+
+:global(.react-datepicker__navigation) {
+  position: absolute;
+  width: 3rem;
+  border: none;
+  text-indent: -999em;
+  mask: url('../../assets/icons/caret-down.svg') no-repeat center center;
```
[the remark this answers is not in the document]
nighto: Thanks for catching it, for whatever reason the relative path is not working, needed to pass the full path starting with `src/`.

mkrause (approved):

PR #49, opened 2024-12-08
Implement Disclosure and Accordion components

mkrause:
This PR implements two new components:

- Disclosure – A collapsible element with a summary title.
- Accordion – A sequence of disclosure components that can be opened either exclusively or non-exclusively.

Screenshot 2024-12-23 at 12 20 12


src/components/containers/Accordion/Accordion.module.scss
```diff
@@ -0,0 +1,22 @@
+/* Copyright (c) Fortanix, Inc.
+|* This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy of
+|* the MPL was not distributed with this file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+@use '../../../styling/defs.scss' as bk;
+
+@layer baklava.components {
+  .bk-accordion {
```
spli02, line 8: When accordion item is opened or closed, the entire accordion width is changed. Is it intentional? (Seems like it is not changed in baklava1, and I feel like the same width looks better.)

https://github.com/user-attachments/assets/4bf3df72-69e9-4bc8-a589-e22e6b14b86e

src/components/containers/Disclosure/Disclosure.stories.tsx
```diff
@@ -0,0 +1,86 @@
...
+    open: true,
+    summaryProps: { className: 'pseudo-focus-visible' },
+    title: 'I should have a focus outline',
+    children: <>When open, the focus outline should have a border radius on the top only.</>,
+  },
+};
+
+export const WithTextOverflow: Story = {
+  args: {
+    title: 'A long title that should cause text overflow with an ellipsis',
```
spli02, line 69: For a long text title, is it better to show the entire title by using tooltip on mouseover?
Screenshot from 2024-12-23 11-12-04

src/components/containers/Disclosure/Disclosure.module.scss
```diff
@@ -0,0 +1,140 @@
...
+        transition-property: block-size, content-visibility;
+        transition-duration: var(--bk-disclosure-transition-duration);
+        transition-behavior: allow-discrete;
+      }
+    }
+    &[open]::details-content {
+      block-size: auto;
+    }
+    
+    // Fallback styling for browsers that don't support `::details-content`
```
spli02, line 114: We don't support transition movement for browsers that don't support `::details-content`?
In firefox, the accordion open/close movement is not smooth like chrome.

In below video, the first test is firefox, the second test is chrome.

https://github.com/user-attachments/assets/6ec1bd52-7d1c-4832-a450-4723cacb3f85

src/components/containers/Disclosure/Disclosure.module.scss
```diff
@@ -0,0 +1,140 @@
...
+      
+      .bk-disclosure__collapse-icon {
+        rotate: -0.5turn;
+      }
+    }
+    
+    // Inner wrapper for content. We cannot give `::details-content` any margin/padding because it would still be
+    // visible (due to `content-visibility: hidden` still taking up space). So we use this as scroller and for spacing.
+    .bk-disclosure__content {
+      overflow-y: scroll;
```
spli02, line 88: Scroll bar is displayed inside of all accordion items, even though it's not scrollable. 
Is it better to hide this scroll bar?
Screenshot from 2024-12-23 10-48-47

src/util/storybook/LayoutDecorator.module.scss
```diff
@@ -0,0 +1,13 @@
+
```
spli02, line 1: Copyright is missing? Also for the following 2 files:
`src/util/storybook/LayoutDecorator.tsx`
`src/util/storybook/LoremIpsum.tsx`

spli02 (changes requested):

src/util/storybook/LayoutDecorator.module.scss
mkrause: Fixed! Thanks

src/components/containers/Disclosure/Disclosure.module.scss
mkrause: Ah yeah, good catch. I was testing this on Mac where the scrollbars aren't shown anyway. Needed to use `overflow-y: auto` instead. Fixed now.

src/components/containers/Disclosure/Disclosure.stories.tsx
mkrause: Right, we should have some better text overflow management. I've created a separate ticket for it: https://github.com/fortanix/baklava/issues/75

I would want to do this with a reusable component/hook, since it will involve some JS logic for overflow detection.

src/components/containers/Disclosure/Disclosure.module.scss
mkrause: Right, I implemented this component using `<details>` rather than inventing my own using `<div>`s and a lot of `aria-` attributes. `<details>` is pretty much perfect for this use case, and has decent accessibility out of the box. The downside though is that Firefox and Safari don't yet support `::details-content` and that means we cannot implement certain things like animating the open/close transition. Safari just landed `::details-content` though so that will be in an upcoming release.

There's a good video on the topic here from Kevin Powell: https://www.youtube.com/watch?v=Vzj3jSUbMtI

My idea here is to make this a progressive enhancement: browsers that support it will get the animation, browsers that don't still get all the working functionality just without the animation.

src/components/containers/Accordion/Accordion.module.scss
mkrause: This is actually an issue with the story, not the component itself. I was working on the `LayoutDecorator` utility component for Storybook, but it doesn't quite work yet. So what's happening is that the story here doesn't have a fixed width defined and instead just shrinks to the max content of the `Accordion`.

I just pushed a workaround. Will revisit later.

src/components/containers/Disclosure/Disclosure.module.scss
spli02: With `overflow-y: auto` change, scroll bar is hidden when accordion item is fully open. 
But while we open or close it, the scroll bar is displayed (please refer the video below.).

https://github.com/user-attachments/assets/6acd2708-845c-455b-bab8-a27c0874ffc8

mkrause: @spli02 You're right that's a bit jarring. I pushed a fix, now the `overflow-y: auto` only gets applied after the transition is complete on open, and it gets removed immediately upon triggering close.

spli02 (approved):

PR #50, opened 2024-12-08
Update icons

mkrause:


spli02:
I'm not that familiar around icon component, but some icons from `AKS` to `Hallucination` on Figma don't seem to be exist under `src/assets/icons_source` folder. Is it fine for now?
Screenshot from 2024-12-09 11-15-53

spli02 (approved):

PR #52, opened 2024-12-09
Refactor top-level Sass code

mkrause:
- Removes usages of legacy @import statements.
- Fixed @layer ordering not being emitted at the top of the build file.


nighto (approved):
LGTM :rocket:

PR #54, opened 2024-12-10
Update icons

mkrause:
- Update icons as per latest Figma export
- Remove now unnecessary renames/skips


mkrause:
TODO:

- [] Fix a few more naming issues in the source icons
- [] Add missing graphics (and add those to the relevant places, like PlaceholderEmpty)

nighto:
Close buttons (cross) are considerably smaller than before, not sure if that's expected https://www.chromatic.com/test?appId=6757628e1f35f97b959b83b4&id=675b7da316bc53f4294facde

mkrause:
> Close buttons (cross) are considerably smaller than before, not sure if that's expected https://www.chromatic.com/test?appId=6757628e1f35f97b959b83b4&id=675b7da316bc53f4294facde

It's expected yeah, I had discussed this with Leif. There are three variants (small, medium, large) of the cross icon in the Figma and we went for the middle one. Whether that's the right size I'm not sure, the whole icon pack really needs some work to get everything visually similar sized.

mkrause:
@nighto It's a good catch though because this affects some components that are currently expecting the cross icon to be a certain size. I'll go through and review those components.

src/assets/icons/cross.svg
mkrause, line 1: Need to review the components that use this to see if there are significant visual regressions due to the icon change

nighto (approved):

PR #65, opened 2024-12-18
Update Figma import script + move generated code to separate files

mkrause:
This PR refactors the `scripts/import.ts` script:

- Add a new command to import primitive color values
- Automatically write the generated code to files in the new `src/styles/generated` directory (rather than the user having to do it manually)
- Also generate TypeScript files to read the tokens if necessary (e.g. to support the `Colors.mdx` Storybook documentation)

TODO:
- [ ] Give the same treatment to the other token types: font sizes, border radii, and sizings
- [ ] Update the components that are using tokens that are now missing in Figma.


scripts/import.ts
```diff
@@ -252,6 +432,7 @@ const printUsage = () => {
     
     Commands:
       - parse-tokens
```
spli02, line 434: `parse-tokens` command was deleted in L288, so should be changed to `import-colors-semantic`?

scripts/import.ts
```diff
@@ -156,15 +335,16 @@ const validateIcon = async (path: string, iconName: string): Promise<IconValidit
       throw new Error(`Expect icon dimensions to be 18x18, found ${width}x${height}`);
     }
     
-    // TODO: accessibility checks? E.g. each icon should have an accessible name.
```
spli02, line 159: If we no longer check for icon name, we don't have to accept `iconName` as the second parameter in `validateIcon` function?

scripts/import.ts
```diff
@@ -47,77 +47,256 @@ const rel = (absolutePath: string): string => {
   return path.relative(process.cwd(), absolutePath);
 };
 
+const addGenerationComment = (fileContents: string) => dedent`
```
spli02, line 50: It's not due to the changes of this PR, but I noticed that there is a warning because `await` is not defined inside of `async` function in L482 and L484. 
await

Seems like there are some ways to fix it: https://stackoverflow.com/questions/66486903/top-level-await-expressions-are-only-allowed-when-the-module-option-is-set-t. We will also fix it?

spli02 (changes requested):

scripts/import.ts
mkrause: This is top-level await, which is supported in Node for a while now. Not sure why TypeScript is complaining about it here, we already have `module` set to "ESNext":

```
// tsconfig.node.json
"module": "ESNext", // The type of file to emit (CommonJS/ESM/etc.)
```

My guess is that in your IDE it's not picking up this tsconfig for some reason.

scripts/import.ts
mkrause: Good catch! Fixed it.

scripts/import.ts
mkrause: Might be useful in the future. But I've removed it for now, can always add it back if needed.

spli02 (approved):

PR #67, opened 2024-12-20
Refactor Input with tags to display tags below

nighto:


src/components/forms/fields/InputFieldWithTags/InputFieldWithTags.stories.tsx
```diff
@@ -49,14 +49,16 @@ export const InputWithTags: Story = {
...
         <InputFieldWithTags
           tags={tags}
           value={inputText}
+          label={'Input with tags'}
           onUpdate={handleUpdate}
           onUpdateTags={handleUpdateTags}
-          placeholder=""
+          placeholder="Placeholder"
         />
+        <p>(enter creates new tags; backspace removes tags)</p>
```
mkrause, line 61: Maybe better to put this in an info alert above the card, rather than inline. Also if you can make the above sentence case (so capital Enter and ending in a period, without the parens).

src/components/forms/fields/InputFieldWithTags/InputFieldWithTags.tsx
```diff
@@ -110,23 +110,20 @@ export const InputFieldWithTags = (props: InputFieldWithTagsProps) => {
...
+        form={formContext.formId}
+        className={cx(cl['bk-input-field-with-tags__control'], inputProps.className)}
+        onChange={onChange}
+        onKeyDown={onKeyDown}
+        value={value}
+      />
+      <div className={cl['bk-input-field-with-tags__tags']}>
         {tags && (
           // biome-ignore lint/suspicious/noArrayIndexKey: no other unique identifier available
           tags.map((tag, idx) => <Tag key={idx} content={tag} onRemove={() => onRemoveTag(idx)}/>)
```
mkrause, line 125: This is more of an issue in the `Tag` component itself, but the close button is not keyboard navigable at the moment. Need to wrap the close icon in a `<Button unstyled>` element and move the `onClick` to an `onPress` on the button.

mkrause (changes requested):

mkrause (approved):

PR #71, opened 2024-12-20
Change hardcoded month to november instead of december

nighto:


mkrause:
Thanks. For clarity: this was done so that we don't get flaky Storybook diffs since it highlights the current date of the month in bold.

mkrause (approved):

PR #74, opened 2024-12-23
Add Storybook test runner for accessibility testing

mkrause:
This PR adds a new Github Actions workflow for [Storybook test runner](https://storybook.js.org/docs/writing-tests/test-runner). This test runner loads up the generated storybook pages in a headless browser (playwright), and then performs some tests. In particular, it currently is configured with [axe-playwright](https://www.npmjs.com/package/axe-playwright) so that we get accessibility violation checking.

Currently there are still a large amount of accessibility violations, so I have made the workflow non-enforced for now, it can instead be used for informative purposes and to track the number of violations.

Screenshot 2024-12-24 at 16 01 01


.storybook/test-runner.ts
```diff
@@ -0,0 +1,31 @@
+import { type TestRunnerConfig, getStoryContext } from '@storybook/test-runner';
```
spli02, line 1: Most files under `.storybook` folder has copyright, so copyright is needed?

spli02 (approved):

.storybook/test-runner.ts
mkrause: Thanks, you're right. I've added it now.

PR #78, opened 2024-12-25
Banner component update

mkrause:
This PR:

- Reworks and expands on the `Banner` component.
- Removes the `Alert` component, which is a duplication.
- Adds new "filled" variants of the success/error/info icons, for use in `Banner`.

Screenshot 2024-12-28 at 22 43 08

It also:

- Fixes an issue with Storybook controls no longer working, a regression that happened with the splitting of `tsconfig.json` files.
- Fixes some issues in the `SegmentedControl` and `Tooltip` components (which I needed for Banner).

TODO:
- [x] Need to get the right icons, the Figma has some icons that are not in the icon pack (e.g. filled success/error icons).


mkrause:
This apparently caused a regression in Chrome v121, which is the version used on Chromatic at the moment:

Screenshot 2024-12-30 at 18 25 29

I cannot reproduce this issue in any modern browser, and just from the screenshots I have not been able to deduce the issue. So this would need some investigation on an actual Chrome v121 instance. On the other hand, Chrome v121 was released in January 2024, so it's almost a year old by now and maybe we can ignore this as an old browser bug.

src/util/reactDomUtil.ts
```diff
@@ -0,0 +1,14 @@
+
```
spli02, line 1: Copyright?

src/util/storybook/storybookUtil.ts
```diff
@@ -0,0 +1,10 @@
+
+export const storyDescription = (description: string) => ({
```
spli02, line 2: 1. `storyDescription` doesn't seem to be used in the project. Is it necessary to define it?
2. If it's necessary, copyright is missing in this file.

spli02 (changes requested):

src/components/containers/Banner/Banner.module.scss
```diff
@@ -7,93 +7,127 @@
...
+            color: var(--bk-banner-color-foreground);
+          }
+          
+          .bk-banner__title__text {
+            @include bk.text-one-line;
+          }
+        }
+        
+        .bk-banner__message--compact {
+          --keep: ;
```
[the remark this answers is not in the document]
mkrause: If a selector has no content it will get stripped by the transpiler. Which breaks CSS modules since we import this selector in the component. We could also remove it entirely from the component, but I think it would be nice to keep these for future use.

src/util/storybook/storybookUtil.ts
mkrause: I was using it for a bit but then find out a docblock can serve the same purpose. Removed this file now.

src/util/reactDomUtil.ts
mkrause: Done

src/components/containers/Banner/Banner.tsx
```diff
@@ -2,125 +2,184 @@
...
- * A wrapper component, intended to easily add some styling to children's `<Button/>`'s. UX expects that such buttons are `<Button variant="tertiary"/>`.
+ * An action that is rendered as just an icon.
  */
-export const BannerAction = (props: BannerActionProps) => {
+export const ActionIcon = ({ tooltip, ...buttonProps }: ActionIconProps) => {
   return (
-    <div className={cl['bk-banner__action']}>
-      {props.children}
-    </div>
+    <TooltipProvider compact tooltip={typeof tooltip !== 'undefined' ? tooltip : buttonProps.label}>
```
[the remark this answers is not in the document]
mkrause: Right, this is sort of on purpose, since the banner itself is in light mode. The banner has a `bk-theme--light` override, so all content within it will be light mode. This seems correct to me, since a dark button on the light banner background would look weird.

For tooltips though: maybe you're right that dark mode would look better for those? Should all tooltips follow the "global" (root) theme maybe? I'll open a ticket for it, think it needs some UX input.

src/components/actions/Button/Button.tsx
```diff
@@ -20,6 +20,9 @@ export type ButtonProps = React.PropsWithChildren<Omit<ComponentProps<'button'>,
   /** Whether this component should be unstyled. */
   unstyled?: undefined | boolean,
   
+  /** Whether to trim this component (strip any spacing around the element). */
+  trimmed?: undefined | boolean,
```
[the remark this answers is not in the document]
mkrause: Yeah, good idea. I've added a story.

spli02 (approved):

PR #85, opened 2025-01-02
Modal components update

mkrause:
This PR:
- Introduces a new `ModalProvider` component that can be used to create controlled modal dialogs.
- Implements a `DialogModal` component which uses `ModalProvider` and `Dialog` to implement various modals (centered, sliding, full screen).
- Implements `useModalWithSubject` for easy sharing of a modal component.
- Implements `useConfirmation` that builds on `useModalWithSubject` but is tailored for confirmation dialogs.
- Adds a `SpinnerModal` component for a modal `Spinner`.
- Removes the `Modal` component, consumers will need to use `DialogModal` instead.

Screenshot 2025-01-05 at 01 16 27

TODO:
- [x] Implement a way for Dialog action buttons to communicate with the Dialog/Modal. Currently in the stories none of the action buttons (like Cancel, Submit) are able to close the modal.
- [ ] Currently modals don't have a minimum size. Investigate if we can set a minimum (and what that should be), without breaking responsive design (i.e. modals should still be able to shrink on small viewports). EDIT: Split out to https://github.com/fortanix/baklava/issues/91
- [x] Investigate Safari bug (see comment below).


mkrause:
Need to investigate a bug in Safari 17.6 where, when opening a submodal, the close button in the parent modal remains focused. When you then press "space", all the modals are suddenly closed although the page remains inert for a bit until the dialogs unmount.

Screenshot 2025-01-05 at 20 53 14

- [ ] Investigate Safari bug.

src/components/containers/Dialog/Dialog.module.scss
```diff
@@ -81,10 +89,16 @@
       
       .bk-dialog__action { --keep: ; }
       .bk-dialog__action--icon { --keep: ; }
-      .bk-dialog__action--button { --keep: ; }
     }
     
     // State: focus
     @include bk.focus-outset;
+    // .bk-dialog__content {
```
spli02, line 96: These comments should be added?

src/components/containers/Dialog/Dialog.stories.tsx
```diff
@@ -28,9 +28,7 @@ export default {
   args: {
     title: 'Dialog title',
     children: <LoremIpsum paragraphs={3}/>,
-    actions: (
-      <Button variant="primary" label="Submit"/>
-    ),
+    actions: <Dialog.SubmitAction/>,
```
spli02, line 31: `Button` component import of L11 can be removed.
`import { Button } from '../../actions/Button/Button.tsx';`

src/components/overlays/ModalProvider/ModalProvider.stories.tsx
```diff
@@ -0,0 +1,67 @@
...
+    children: ({ activate }) => <Button variant="primary" label="Open modal" onPress={activate}/>,
+    dialog: ({ close, dialogProps }) => <dialog {...dialogProps}>Content</dialog>,
+  },
+};
+
+const ModalProviderAutoOpen = (props: React.ComponentProps<typeof ModalProvider>) => {
+  const ref = ModalProvider.useRef(null);
+  
+  // biome-ignore lint/correctness/useExhaustiveDependencies: want to only trigger this once
+    React.useEffect(() => {
```
spli02, line 52: It's a minor thing, but one indent before `React.useEffect(() => {` can be removed.

spli02 (approved):

PR #86, opened 2025-01-03
Fix fractional px value rounding issues

mkrause:
Due to the fluid typography we use, most lengths end up being some fraction of a px. Browsers will round this value, but this can cause some visual glitches. For example:

image

In `Dialog`, see the very bottom of the dialog where one pixel row of text is bleeding through, despite the footer being positioned with `bottom: 0`.

image

Here in `Accordion`, the 1px horizontal border on “Title 2” is hidden due to overlap from the “Title 3” having shifted upwards slightly. This is despite all these elements being in a basic flex box with a `gap: 0`.

Nowadays, browsers support the CSS `round()` function. This is baseline as of May 2024. Using this function we can round our values to the nearest pixel.

For some thing slightly more advanced, we can even make use of the `calc-size` function to round a width/height of `auto`. (But `calc-size` is not yet well enough supported, we can't use it.) See [here](https://css-tip.com/pixel-rounding). This doesn't seem necessary though, just the `round()` on the `rem` values seems sufficient to fix all the rounding issues I've come across.

### DIFF
--- a/src/components/containers/Dialog/Dialog.module.scss
+++ b/src/components/containers/Dialog/Dialog.module.scss
@@ -35,7 +35,7 @@
       top: 0;
       z-index: 1;
       
-      padding-block: var(--bk-dialog-padding-block) (bk.$spacing-7 / 2);
+      padding-block: var(--bk-dialog-padding-block) calc(bk.$spacing-7 / 2);
       padding-inline: var(--bk-dialog-padding-inline);
       background: var(--bk-dialog-background-color);
       
@@ -58,7 +58,7 @@
     .bk-dialog__content {
       flex: 1;
       
-      margin-block: (bk.$spacing-7 / 2) (bk.$spacing-7 / 2);
+      margin-block: calc(bk.$spacing-7 / 2) calc(bk.$spacing-7 / 2);
       padding-inline: var(--bk-dialog-padding-inline);
     }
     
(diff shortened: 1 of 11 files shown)


src/components/containers/Dialog/Dialog.module.scss
```diff
@@ -35,7 +35,7 @@
       top: 0;
       z-index: 1;
       
-      padding-block: var(--bk-dialog-padding-block) (bk.$spacing-7 / 2);
+      padding-block: var(--bk-dialog-padding-block) calc(bk.$spacing-7 / 2);
```
spli02, line 38: I'm not sure if I understand the issue correctly, but the most bottom of text is still not fully displayed on my environment in dialog. Is it expected?
Screenshot from 2025-01-06 16-20-43

spli02 (changes requested):

src/components/containers/Dialog/Dialog.module.scss
mkrause: @spli02 If you look at the top screenshot in the issue description, notice there is a very thin 1px row of text visible below the footer. That 1px row is caused a rounding issue. In your screenshot there is no such 1px row.

To reproduce this issue you may need to resize your browser a bit, the rounding issue only happened on specific screen sizes.

spli02: I see. I misunderstood the issue. I'll approve the PR.

spli02 (approved):

PR #105, opened 2025-01-15
Adds missing border to Tag (InputWithTags)

nighto:
Also removes one unused CSS call that was leaking as 'undefined' to the class list

### DIFF
--- a/src/components/text/Tag/Tag.module.scss
+++ b/src/components/text/Tag/Tag.module.scss
@@ -12,6 +12,7 @@
     
     padding: bk.$size-2 bk.$spacing-2;
     border-radius: bk.$size-2;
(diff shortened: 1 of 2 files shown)


src/components/text/Tag/Tag.module.scss
```diff
@@ -12,6 +12,7 @@
     
     padding: bk.$size-2 bk.$spacing-2;
     border-radius: bk.$size-2;
+    border: 1px solid bk.$theme-tag-border-default;
```
mkrause, line 15: Can you add the rem-from-px mixin here?

nighto: Sure

mkrause (approved):

PR #107, opened 2025-01-15
New component: TextArea

nighto:


src/components/forms/controls/Textarea/Textarea.module.scss
```diff
@@ -0,0 +1,38 @@
...
+
+    --bk-textarea-border-color: #{bk.$theme-text-area-border-default};
+
+    border: 1px solid var(--bk-textarea-border-color);
+    border-radius: bk.$radius-1;
+    color: bk.$theme-text-area-text-default;
+    padding: bk.$spacing-3;
+
+    &:focus, &.pseudo-focused {
+      outline: none !important;
```
mkrause, line 19: You can use `@include bk.focus-hidden` to hide it instead (note: should be at the top level, not inside a `&:focus`).

app/lib.ts
```diff
@@ -38,6 +38,7 @@ export { Radio } from '../src/components/forms/controls/Radio/Radio.tsx';
 export { SegmentedControl } from '../src/components/forms/controls/SegmentedControl/SegmentedControl.tsx';
 export { Select } from '../src/components/forms/controls/Select/Select.tsx';
 export { Switch } from '../src/components/forms/controls/Switch/Switch.tsx';
+export { Textarea } from '../src/components/forms/controls/Textarea/Textarea.tsx';
```
mkrause, line 41: Let's name the component `TextArea` (and `text-area` in kebab).

src/components/forms/controls/Textarea/Textarea.tsx
```diff
@@ -0,0 +1,46 @@
...
+export type TextareaProps = Omit<ComponentProps<'textarea'>, 'type'> & {
+  /** Whether this component should be unstyled. */
+  unstyled?: undefined | boolean,
+
+  /** 
+   * Whether the textarea should resize automatically, with `field-sizing: content`.
+   * Note that browser support is still somewhat limited:
+   * https://developer.mozilla.org/en-US/docs/Web/CSS/field-sizing
+   */
+  automaticResize?: undefined | boolean,
```
mkrause, line 21: I think this should have the option to only resize vertically or horizontally or both. Usually we'd only want the text area to grow vertically, not horizontally.

src/components/forms/controls/Textarea/Textarea.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  invalid?: undefined | boolean,
```
mkrause, line 23: Can you add a doc comment here?

src/components/forms/controls/Textarea/Textarea.module.scss
```diff
@@ -0,0 +1,38 @@
...
+
+@use '../../../../styling/defs.scss' as bk;
+
+@layer baklava.components {
+  .bk-textarea {
+    @include bk.component-base(bk-textarea);
+
+    --bk-textarea-border-color: #{bk.$theme-text-area-border-default};
+
+    border: 1px solid var(--bk-textarea-border-color);
```
mkrause, line 13: Use `bk.rem-from-px` here

mkrause (changes requested):

src/components/forms/controls/Textarea/Textarea.tsx
nighto: I am simply forwarding the CSS `field-sizing: content` which does both. The way to resize only vertically would be by setting a fixed `width` to it. Nevertheless I created a separate story exploring that option.

mkrause (approved):

Issue #97, opened 2025-01-07
Firefox: bk-body-text is broken


mkrause:
The `.bk-body-text` class is broken in Firefox currently, need to investigate how to do this without `@scope` support.

PR #117, opened 2025-01-23
Update global styling system (`.bk-prose`, `.bk-no-globals`, + more)

### LINKED ISSUE #61: Enforce no !important in stylelint
Currently there's a decent number of `!important` calls. Need to go through these and see if we can refactor them out or if they're necessary (and thus need a stylelint ignore). Once done, enable the `declaration-no-important` rule.

mkrause:
Resolves #61 , #97, #158.

This PR:
- Renames `.bk-body-text` to `.bk-prose`. It also updates the implementation of `.bk-prose` so that it has a fallback for Firefox which doesn't yet support `@scope` (#97).
- Implements the ability to opt out of global styling through `<html class="bk-no-globals">`. This is added for consumers that need to control the root styling, for example if the application needs to set a different `rem` value (e.g. the 1rem = 62.5% = 10px trick). See also #159.
- Remove `normalize.css` since the `reset.css` makes this mostly useless (but we can incorporate some of the normalize code into `bk-prose`.
- Fixed an issue where some code was referencing `--bk-layout-header-height` rather than the (renamed) `--bk-app-layout-header-height`.
- Removes the remaining unnecessary uses of `!important` and enforces its disuse through stylelint (#61).
- Fixes the Storybook light/dark switching issue in Safari (#158).
- Some minor `SegmentedControl` styling fixes when it is used in a flexbox context.
- Silences the [Sass `mixed-decls` deprecations](https://sass-lang.com/documentation/breaking-changes/mixed-decls/) for now, we need to be wary of it but wrapping all declarations in `& {}` is a non-starter and hard to undo. Once [this ticket](https://github.com/sass/sass/issues/3960) is implemented we should opt in to the new behavior.

TODO:

- [x] Instead of a `.bk-root` opt-in, consider adding a `.bk-no-globals` opt out. We can remove the current `.bk-root` system altogether and just make the `root-styling` mixin opt out, plus add something like the `isolate` mixin that would be the root element on Baklava UI (in `BaklavaProvider`?)
- [x] Before merging, undo the default `.bk-root` change, and disable the `TEMP` code in `BaklavaProvider`.
- [x] Fix light/dark mode switching being broken when the class set is set on `:root` (e.g. `bk-theme-light` on `<html>` when user preference is dark mode no longer switches to light).
- [ ] Update `.bk-prose` default element styles?


nighto (approved):
LGTM :rocket:

PR #136, opened 2025-02-11
Split Checkbox component into a boolean Checkbox and a tri-state CheckboxTri (which also allows indeterminate state)

mkrause:
Previously, we had one component `Checkbox` which would allow either on/off/indeterminate. However, the implementation had some bugs in certain edge cases. I tested both uncontrolled and controlled scenarios and these now work as expected. I decided to split out the two into a `Checkbox` component and a `CheckboxTri` (with indeterminate). For one because it makes the types more precise (no need to ignore the `indeterminate` state in `onUpdate`, but also it means that regular checkboxes don't need the additional complexity.

TODO:
- [x] Fix the entry animation for `:indeterminate`.
- [x] The color for disabled checkmarks is not per design. The checkmark is a `data:` URL SVG but that doesn't allow a dynamic color.
- [x] Check on the missing designs for Switch component.

### DIFF
--- a/src/components/forms/controls/Checkbox/CheckboxTri.stories.tsx
+++ b/src/components/forms/controls/Checkbox/CheckboxTri.stories.tsx
@@ -72,6 +72,7 @@ const CheckboxTriControlled = (args: CheckboxTriArgs) => {
     <div style={{ textAlign: 'center' }}>
       <CheckboxTri
         {...args}
+        defaultChecked={undefined} // `defaultChecked` must be `undefined` for controlled components
         checked={checked}
         onMouseDown={event => { shiftKeyRef.current = event.shiftKey; }} // For click events
         onKeyDown={event => { shiftKeyRef.current = event.shiftKey; }} // For key events (spacebar)
--- a/src/components/forms/controls/Checkbox/CheckboxTri.stories.tsx
+++ b/src/components/forms/controls/Checkbox/CheckboxTri.stories.tsx
@@ -6,14 +6,14 @@ import type { Meta, StoryObj } from '@storybook/react';
 
 import * as React from 'react';
 
-import { CheckboxTernary } from './CheckboxTernary.tsx';
+import { CheckboxTri } from './CheckboxTri.tsx';
 
 
-type CheckboxTernaryArgs = React.ComponentProps<typeof CheckboxTernary>;
-type Story = StoryObj<CheckboxTernaryArgs>;
+type CheckboxTriArgs = React.ComponentProps<typeof CheckboxTri>;
+type Story = StoryObj<CheckboxTriArgs>;
 
 export default {
-  component: CheckboxTernary,
+  component: CheckboxTri,
   parameters: {
     layout: 'centered',
   },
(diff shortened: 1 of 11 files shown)


nighto:
Codewise looks nice, but I noticed a difference in the layout, was it intentional? The checkmark and the dash mark (for indeterminate) are slightly bigger than before.

image

nighto (changes requested):
Just double checked on the Figma and it looks as it was before, so I will mark this one as not approved for now.

image

src/components/forms/controls/Checkbox/CheckboxTri.stories.tsx
```diff
@@ -0,0 +1,94 @@
...
+  // XXX we could maybe turn this shift-click into a feature:
+  // `<CheckboxTri allowUserAction="none | shift"/>`
+  return (
+    <div style={{ textAlign: 'center' }}>
+      <CheckboxTri
+        {...args}
+        defaultChecked={undefined} // `defaultChecked` must be `undefined` for controlled components
+        checked={checked}
+        onMouseDown={event => { shiftKeyRef.current = event.shiftKey; }} // For click events
+        onKeyDown={event => { shiftKeyRef.current = event.shiftKey; }} // For key events (spacebar)
```
nighto, line 78: spacebar? (probably was a left over from some other example?)

src/components/forms/controls/Checkbox/CheckboxTri.stories.tsx
```diff
@@ -0,0 +1,94 @@
...
+    disabled: true,
+    defaultChecked: 'indeterminate',
+  },
+};
+
+const CheckboxTriControlled = (args: CheckboxTriArgs) => {
+  const shiftKeyRef = React.useRef(false);
+  const [checked, setChecked] = React.useState<CheckboxTriArgs['checked']>(args.defaultChecked ?? false);
+  
+  // XXX we could maybe turn this shift-click into a feature:
```
nighto, line 69: XXX = Left over from some personal comment? Maybe just replace if for TODO

nighto:
Also, maybe we could have a ticket (for a next moment) to make a `CheckboxTriField` and `CheckboxTriGroup` just for the sake of completion, because now trying to set those as indeterminate would yield a type error.

src/components/forms/controls/Checkbox/CheckboxTri.stories.tsx
mkrause: "XXX" comments are a fairly common convention in code: https://stackoverflow.com/questions/1452934/what-is-the-meaning-of-xxx-in-code-comments

Could be a "TODO" also, but this isn't really something that needs to be done/fixed, more of a "hey here's a thought".

nighto:
TIL, thanks

mkrause:
> Also, maybe we could have a ticket (for a next moment) to make a `CheckboxTriField` and `CheckboxTriGroup` just for the sake of completion, because now trying to set those as indeterminate would yield a type error.

Good point. Done: https://github.com/fortanix/baklava/issues/145

src/components/forms/controls/Checkbox/CheckboxTri.stories.tsx
nighto: Oh, of course, thanks

mkrause:
Tweaked the checkmark/dash a bit to match the design spec.

Issue #133, opened 2025-02-10
Localization – Support alternative `direction` and `writing-mode` values


mkrause:
We have already switched most of the codebase over to use [CSS logical properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_logical_properties_and_values), but a few stragglers remain. We should convert those, and then configure stylelint to enforce logical properties:

https://github.com/csstools/stylelint-use-logical

Also add some example stories with `direction: rtl`, and possibly `writing-mode: vertical-rl`.

PR #144, opened 2025-02-13
Enforce logical properties (partial, still excluding width/height)

mkrause:
This PR partially addresses #133. Most logical properties have been converted, with the exception of width/height and related properties like max-width and such. It also starts enforcing this through stylelint.

### DIFF
--- a/src/components/containers/Banner/Banner.stories.tsx
+++ b/src/components/containers/Banner/Banner.stories.tsx
@@ -162,7 +162,7 @@ export const BannerWithThemedContent: Story = {
           The following components should always have a light theme, even in dark mode:
         </p>
         <div style={{ display: 'flex', gap: '2ch', marginTop: '1lh' }}>
-          <Button nonactive kind="primary" onPress={() => alert('clicked')}>Button</Button>
+          <Button nonactive kind="primary" onPress={() => { notify.info('Clicked'); }}>Button</Button>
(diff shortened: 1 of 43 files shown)


src/components/containers/Banner/Banner.stories.tsx
```diff
@@ -64,9 +65,9 @@ export default {
 
 
 const ExampleActionButton = () =>
-  <Banner.ActionButton onPress={() => alert('clicked')}>Button</Banner.ActionButton>;
+  <Banner.ActionButton onPress={() => { notify.info('Clicked'); }}>Button</Banner.ActionButton>;
 const ExampleActionIcon = () =>
-  <Banner.ActionIcon icon="copy" label="Copy" onPress={() => alert('clicked')}/>;
+  <Banner.ActionIcon icon="copy" label="Copy" onPress={() => { notify.info('Clicked'); }}/>;
```
spli02, line 70: `alert` is still used in L165, is it fine?
`<Button nonactive kind="primary" onPress={() => alert('clicked')}>Button</Button>`

spli02 (approved):

src/components/containers/Banner/Banner.stories.tsx
mkrause: Good catch, updated it.

PR #150, opened 2025-02-14
Implement `aside` for `Dialog` + `DialogLayout`

nighto:


nighto:
Double check inner title and padding bottom

nighto:
Move to Layouts section on Storybook

src/layouts/DialogLayout/DialogLayout.stories.tsx
```diff
@@ -0,0 +1,131 @@
...
+                steps={[
+                  { stepKey: '1', title: 'Select Cloud Provider' },
+                  { stepKey: '2', title: 'Setup Cloud Connection' },
+                  { stepKey: '3', title: 'Set Up Subscriptions' },
+                  { stepKey: '4', title: '4th Option Title' },
+                ]}
+                activeKey="1"
+                onSwitch={() => {}}
+              />
+              {/* TODO: Maybe this should be a separate component? */}
```
mkrause, line 57: Probably not a separate component. Could be a mixin. Or you could use `bk-prose` which has styling for `hr`

mkrause (changes requested):

src/components/forms/fields/FieldLayout/FieldLayout.module.scss
```diff
@@ -0,0 +1,19 @@
...
+|* the MPL was not distributed with this file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+@use '../../../../styling/defs.scss' as bk;
+
+@layer baklava.components {
+  .bk-field-layout {
+    @include bk.component-base(bk-field-layout);
+
+    &.bk-field-layout--small {
+      inline-size: calc(bk.$spacing-17 + bk.$spacing-6); // 244px
```
mkrause, line 12: Let's just replace this with `rem-from-px(244)`. In the future we probably need a separate set of variables for this. `spacing-` variables could be limited to just gaps/padding/margins etc.

src/components/graphics/Icon/Icon.tsx
```diff
@@ -21,6 +21,16 @@ export type Decoration = (
   | { type: 'background-circle' }
 );
 
+type EventProps = {
+  event: 'warning',
+};
+const Event = (props: EventProps) => {
+  const icon: IconName = props.event === 'warning' ? 'warning-filled' : props.event;
```
mkrause, line 28: Let's make it so that this only works for the valid types of the `event` prop. Currently it has `event: 'warning'`, so if `event` is not `warning` then this component should not render.

src/layouts/DialogLayout/DialogLayout.stories.tsx
```diff
@@ -0,0 +1,133 @@
...
+              <hr/>
+              <div style={{display: 'flex', flexDirection: 'row', gap: '4px', alignItems: 'flex-start'}}>
+                <p><Icon icon="info"/></p>
+                <p><small>Step-by-Step Guide to Connect to Azure Cloud Provider</small></p>
```
mkrause, line 62: Let's not use the `<small>` element, since it doesn't really add anything in terms of semantics.

src/layouts/DialogLayout/DialogLayout.module.scss
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    small {
+      color: bk.$theme-text-small-text-subtle;
+      font-size: bk.$font-size-s;
+    }
+
+    button {
+      font-size: bk.$font-size-s;
+    }
```
mkrause, line 38: These three CSS rules above are too generic. For example, this would style _any_ button in a dialog layout to have a small font.

- The `small` element should be moved into the `__aside` selector, and replaced with a class name, something like `.bk-dialog-layout__content__aside__hint`.
- For `hr`: you can probably just remove this entirely and give the `__hint` above a border-top.
- For `button`: I'm actually not sure what this is needed for. Maybe just remove it entirely?

mkrause (changes requested):

src/layouts/DialogLayout/DialogLayout.stories.tsx
```diff
@@ -55,12 +55,10 @@ export const DialogOnboarding: Story = {
                 activeKey="1"
                 onSwitch={() => {}}
               />
-              {/* TODO: Maybe this should be a separate component? */}
-              <hr/>
-              <div style={{display: 'flex', flexDirection: 'row', gap: '4px', alignItems: 'flex-start'}}>
-                <p><Icon icon="info"/></p>
-                <p><small>Step-by-Step Guide to Connect to Azure Cloud Provider</small></p>
-              </div>
+              <p className={cl['bk-dialog-layout__content__aside__hint']}>
```
mkrause, line 58: Instead of using `DialogLayoutClassNames` (which is not exported), we can add a `DialogLayout.Hint` subcomponent.

mkrause (changes requested):

Issue #145, opened 2025-02-13
Update `CheckboxField` to allow using a `CheckboxTri`?


mkrause:
See comment here: https://github.com/fortanix/baklava/pull/136#issuecomment-2656410255

mkrause:
For a `CheckboxTriGroup`, what would the type of the state be? For `CheckboxGroup`, this is `Set<CheckboxKey>`, where if the checkbox is `checked` then it's in the `Set` otherwise not. But for `CheckboxTri` this wouldn't work, since you can't have "half" a set membership.

Instead, we might need to have something like a `Map<CheckKey, boolean | 'undefined'>`.

PR #151, opened 2025-02-17
Implement `Label` and `FieldSet` components + form control updates

mkrause:
This PR adds two new components: `Label` and `FieldSet`, corresponding to the native HTML `<label>` and `<fieldset>` elements. Additionally, it turns `RadioGroup` and `CheckboxGroup` into proper form controls, with (1) their own state management, (2) both uncontrolled and controlled usage, and (3) automatically communicating state changes and props like `name` and `form` to the subcomponents through context.

Follow ups:
- `SegmentedControl`: add story for controlled usage (plus within a form?)
- `CheckboxTriGroup`: #145


app/lib.ts
```diff
@@ -31,29 +31,33 @@ export { Accordion } from '../src/components/containers/Accordion/Accordion.tsx'
 export { Form } from '../src/components/forms/context/Form/Form.tsx';
 export { SubmitButton } from '../src/components/forms/context/SubmitButton/SubmitButton.tsx';
 
+// Forms > Common
+export { Label } from '../src/components/forms/common/Label/Label.tsx';
```
nighto, line 35: minor nitpick: would be nice to maintain alphabetical order here and in the next section

src/components/forms/common/FieldSet/FieldSet.module.scss
```diff
@@ -0,0 +1,35 @@
...
+      
+      @include bk.text-one-line($include-padding: false); // Padding fix not needed thanks to `overflow: visible`
+    }
+    
+    .bk-field-set__content {
+      // Workaround for vertical wrapping (see "Vertical With Wrap" story). To make vertical wrapping work, the inner
+      // `<div>`'s height must be contained within the `<fieldset>`. This can be done easily with flex/grid, but that
+      // doesn't work well with `<fieldset>`. So instead, we calculate the `max-height` as 100% minus the height of
+      // the `<legend>` (which is contrained to one line thanks to the `text-one-line` mixin.
+      max-height: calc(100% - 1lh - $legend-gap);
```
nighto, line 32: nice trick

app/lib.ts
mkrause: I'm trying to match this with the storybook ordering instead (so as in `preview.tsx`)

PR #155, opened 2025-02-20
Migrate width/height to CSS logical properties

### LINKED ISSUE #133: Localization – Support alternative `direction` and `writing-mode` values

mkrause:
Resolves #133.

### DIFF
--- a/src/components/forms/controls/RadioGroup/RadioGroup.stories.tsx
+++ b/src/components/forms/controls/RadioGroup/RadioGroup.stories.tsx
@@ -135,8 +135,15 @@ export const RadioGroupDirectionRtl: Story = {
 
 export const RadioGroupWritingModeVertical: Story = {
   args: {
-    style: { overflow: 'hidden', resize: 'vertical', padding: 5, inlineSize: 220, blockSize: 140, writingMode: 'vertical-rl' },
-    label: '色を選んでください',
+    style: {
+      overflow: 'hidden',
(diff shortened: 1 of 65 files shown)


src/components/forms/controls/RadioGroup/RadioGroup.stories.tsx
```diff
@@ -118,3 +118,31 @@ export const RadioGroupInForm: Story = {
...
+        <RadioGroup.Button radioKey="green" label="أخضر"/>
+        <RadioGroup.Button radioKey="blue" label="أزرق"/>
+      </>
+    ),
+  },
+};
+
+export const RadioGroupWritingModeVertical: Story = {
+  args: {
+    style: { overflow: 'hidden', resize: 'vertical', padding: 5, inlineSize: 220, blockSize: 140, writingMode: 'vertical-rl' },
```
spli02, line 138: L138 looks too long to be one line, so it's better to format it?

spli02 (approved):

PR #173, opened 2025-04-20
Replace all instances of Dropdown with ListBox, and remove the Dropdown component

mkrause:


src/components/forms/controls/ListBox/ListBox.tsx
```diff
@@ -198,6 +218,9 @@ export const Action = (props: ActionProps) => {
   );
 };
 
+export const FooterAction = (props: Omit<ActionProps, 'sticky'>) => {
+  return <Action {...props} sticky="end"/>;
```
spli02, line 222: I've noticed that there is the same issue as https://github.com/fortanix/baklava/pull/167#discussion_r2048670950 in this branch. (Some text is visible under sticky action.)
Screenshot from 2025-05-06 14-02-19

spli02, line 222: I've noticed that there is the same issue as https://github.com/fortanix/baklava/pull/167#discussion_r2048670950 in this branch. (Some text is visible under sticky action.)
Screenshot from 2025-05-06 14-02-19

spli02 (changes requested):

src/components/forms/controls/ListBox/ListBox.tsx
mkrause: Thanks @spli02, this is fixed now.

spli02 (approved):

PR #180, opened 2025-04-30
introduce spacing variables 19 to 22

nighto:


mkrause:
@nighto Are these from the Figma file? The Figma file should be the source of truth for this list.

If they're not, then please add them as `$size-` tokens.

PR #212, opened 2025-05-17
(#131) Implement `PublicLayout`

### LINKED ISSUE #131: Implement `PublicLayout`
Similar to `AppLayout`, we should implement a set of layout components for the publicly accessible pages like login, signup, email recovery.

mkrause:
Resolves #131

This PR:
- Introduces a new `PublicLayout` component to render the public-facing UI for unauthenticated users.
- Renames `Logo` to `FortanixLogo` to make it more specific. The old `Logo` export has been kept around as an alias for now.
- Adds the ability for `DropdownMenuProvider` items to close the menu through a render prop.

Screenshot 2025-05-18 at 00 22 39

TODO:
- [x] Fix `min-height` rendering issue in Firefox.
- [x] Fix logo rendering issue in Safari.


src/layouts/PublicLayout/PublicLayout.tsx
```diff
@@ -0,0 +1,132 @@
...
+  return (
+    <figure
+      {...propsRest}
+      className={cx(
+        cl['bk-fortanix-armor-logo'],
+        { [cl['bk-fortanix-armor-logo--stacked']]: stacked },
+        propsRest.className,
+    )}
+    >
+      {/* <FortanixLogo/> */}
```
nighto, line 34: Any specific reason not to use the `<FortanixLogo/>`? I assume it's due to the more horizontal nature of the image, with Armor written on the side, instead of the bottom. Eventually that could be a variation of said component. Not a deal breaker, of course.

mkrause: It's blocked by this issue: https://github.com/fortanix/baklava/issues/213

Basically we can't really scale the component at the moment, I had been using `zoom` as a workaround but turns out that doesn't work well in Safari. So for now I'm just using the SVG.

nighto (approved):

src/layouts/PublicLayout/PublicLayout.tsx
nighto: Oh I see, thanks :)

PR #233, opened 2025-05-20
Add disabled state for steps in Stepper component

### LINKED ISSUE #226: Support `disabled` for step property in `Stepper` component
Currently, users can click on any step, and the `Stepper` component doesn't provide a way to disable individual steps.
Ideally, the `step` prop (as defined below) should support `disabled` property, preventing users from clicking on disabled steps.

This is because, in some cases, users shouldn't be allowed to skip ahead - like jumping from Step 1 directly to Step 4 without completing filling form.

```
export type Step = {
  stepKey: string,
  title: React.ReactNode,
  className?: ClassNameArgument,
  hide?: boolean,
  isOptional?: boolean,
};
```

mkrause:
Resolves #226

TODO:
- [ ] What should the disabled step look like once you've completed it? (i.e. progressed beyond it)

### DIFF
--- a/src/components/containers/Card/Card.stories.tsx
+++ b/src/components/containers/Card/Card.stories.tsx
@@ -85,3 +85,24 @@ export const CardGrid: Story = {
     </div>
   ),
 };
+
+export const CardNested: Story = {
+  decorators: [Story => <LayoutDecorator size="small"><Story/></LayoutDecorator>],
+  args: {
+    children: (
+      <>
+        <Card.Heading>A parent card</Card.Heading>
+        <div style={{ display: 'grid', gridTemplateColumns: 'repeat(2, 1fr)' }}>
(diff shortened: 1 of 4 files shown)


src/components/containers/Card/Card.stories.tsx
```diff
@@ -85,3 +85,24 @@ export const CardGrid: Story = {
     </div>
   ),
 };
+
+export const CardNested: Story = {
```
spli02, line 89: `CardNested` change is necessary for this PR?

spli02 (approved):

src/components/containers/Card/Card.stories.tsx
mkrause: It's unrelated, just a minor story addition based on discussion with @arn-craft

PR #234, opened 2025-05-20
DropdownMenuProvider (and thus Select) does not support a default selected state when controlled

mkrause:


mkrause:
This has been solved in a different way for now in https://github.com/fortanix/baklava/pull/256

In that PR, we introduced a `formatItemLabel` prop so that the user can provide their own label formatting. If they don't specify this, then for the first render the `itemKey` will be used as the label instead.

In the future, we may want to revisit this and introduce automatic label formatting based on the `Option` element content. But this will require some communication from the `ListBox` up to the `MenuProvider`. It would be async, so somehow we'd still need to solve the intermediate value before the item actually renders. Plus, there's the challenge of virtual items (through there we could just enforce the `formatItemLabel`?)

PR #322, opened 2025-06-03
Add a table text-cell wrapper with `display: flex`

mkrause:
Currently, adding a `TextLine` component directly in a table cell (`<td>`) doesn't work, since the `<td>` is not a flex parent. This PR adds a flex container parent for all table cells by default.

### DIFF
--- a/src/components/text/TextLine/TextLine.tsx
+++ b/src/components/text/TextLine/TextLine.tsx
@@ -46,7 +46,7 @@ export const TextLine = (props: TextLineProps) => {
   
   if (showOverflowTooltip) {
     return (
-      <TooltipProvider tooltip={renderTooltip}>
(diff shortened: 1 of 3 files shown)


arn-craft (approved):

src/components/text/TextLine/TextLine.tsx
```diff
@@ -46,7 +46,7 @@ export const TextLine = (props: TextLineProps) => {
   
   if (showOverflowTooltip) {
     return (
-      <TooltipProvider tooltip={renderTooltip}>
+      <TooltipProvider tooltip={renderTooltip} triggerAction="click">
```
arn-craft: triggerAction="click" needs to be removed

PR #360, opened 2025-06-10
PageLayout variations

nighto:
Added individual components directly to the AppLayout stories because it made more sense to visualize them.

Includes:
- `PageHeader`, with a `title` and `children` with action buttons on the right

TODO:
- other variations

### DIFF
--- a/src/layouts/AppLayout/AppLayout.stories.tsx
+++ b/src/layouts/AppLayout/AppLayout.stories.tsx
@@ -8,23 +8,25 @@ import type { Meta, StoryObj } from '@storybook/react';
 import { DummyBkLinkUnstyled, DummyBkLinkWithNotify } from '../../util/storybook/StorybookLink.tsx';
 
 import { notify } from '../../components/overlays/ToastProvider/ToastProvider.tsx';
-import { OverflowTester } from '../../util/storybook/OverflowTester.tsx';
 import { Button } from '../../components/actions/Button/Button.tsx';
-import { Panel } from '../../components/containers/Panel/Panel.tsx';
 import { DialogModal } from '../../components/overlays/DialogModal/DialogModal.tsx';
-
-import { Header } from './Header/Header.tsx';
-import { Sidebar } from './Sidebar/Sidebar.tsx';
 import { FortanixLogo } from '../../fortanix/FortanixLogo/FortanixLogo.tsx';
-import { Nav } from './Nav/Nav.tsx';
+import { LoremIpsum } from '../../util/storybook/LoremIpsum.tsx';
+import { OverflowTester } from '../../util/storybook/OverflowTester.tsx';
+import { Panel } from '../../components/containers/Panel/Panel.tsx';
+import { Select } from '../../components/forms/controls/Select/Select.tsx';
+import { Tag } from '../../components/text/Tag/Tag.tsx';
+
+import { AppLayout } from './AppLayout.tsx';
+import { Breadcrumbs } from './Breadcrumbs/Breadcrumbs.tsx';
+import { Header } from './Header/Header.tsx';
 import { UserMenu } from './Header/UserMenu.tsx';
 import { SolutionSelector } from './Header/SolutionSelector.tsx';
 import { AccountSelector } from './Header/AccountSelector.tsx';
 import { SysadminSwitcher } from './Header/SysadminSwitcher.tsx';
-import { Breadcrumbs } from './Breadcrumbs/Breadcrumbs.tsx';
-import { AppLayout } from './AppLayout.tsx';
-import { Select } from '../../components/forms/controls/Select/Select.tsx';
-import { Tag } from '../../components/text/Tag/Tag.tsx';
+import { Nav } from './Nav/Nav.tsx';
+import { PageHeader } from './PageHeader/PageHeader.tsx';
+import { Sidebar } from './Sidebar/Sidebar.tsx';
 
 
 type AppLayoutArgs = React.ComponentProps<typeof AppLayout>;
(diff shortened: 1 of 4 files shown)


mkrause:
The heading is not center aligned currently (relative to the actions)

Screenshot 2025-06-14 at 15 15 08

mkrause:
Missing a gap between the title and actions:

Screenshot 2025-06-14 at 15 24 37

src/layouts/PageLayout/PageHeader/PageHeader.module.scss
```diff
@@ -0,0 +1,26 @@
...
+    
+    .bk-page-header__actions {
+      display: flex;
+      flex-direction: row;
+      gap: bk.$spacing-7;
+    }
+    
+    .bk-page-header__select > div {
+      border-block-end-width: 0;
+    }
```
mkrause, line 24: Removing the line here breaks focus/active states (e.g. you can't see focus because that's done by making the line blue).

But also, I think these specs might be wrong/outdated. Let's keep the `Select` with its basic styling for now.

src/layouts/PageLayout/PageLayout.stories.tsx
```diff
@@ -0,0 +1,153 @@
...
+    ),
+  },
+};
+
+export const PageLayoutTabs: Story = {
+  args: {
+    children: (
+      <>
+        {header1}
+        {tabs1}
```
mkrause, line 149: Let's add the content area (panel) in this story also

src/layouts/PageLayout/PageLayout.stories.tsx
```diff
@@ -0,0 +1,153 @@
...
+    children: (
+      <>
+        {header1}
+        {body}
+      </>
+    ),
+  },
+};
+
+export const PageLayoutSelect: Story = {
```
mkrause, line 133: Instead of "Select", let's use the term "scope", or maybe "scope switcher". It's meant to be a control where you can see/change your current "scope" (like a project, connection, etc.)

src/layouts/PageLayout/PageLayout.module.scss
```diff
@@ -0,0 +1,23 @@
...
+
+@layer baklava.components {
+  .bk-page-layout {
+    @include bk.component-base(bk-page-layout);
+
+    display: flex;
+    flex-direction: column;
+    gap: bk.$spacing-7;
+
+    > [role="tablist"] {
```
mkrause, line 15: This could be fragile. Maybe add a container component, like:

```
<PageLayout>
  <PageLayout.Nav>
    <Tabs>...</Tabs>
  </PageLayout.Nav>
</PageLayout>
```

I'm calling it "Nav" here so that it could be used by other navigation components as well in the future.

src/layouts/PageLayout/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,63 @@
...
+            { [cl['bk-page-header__actions']]: !unstyled },
+          )}>
+            {children}
+          </div>
+        )}
+      </header>
+    );
+  },
+  {},
+);
```
mkrause, line 63: Should we inline this into the `PageLayout.tsx` file maybe?

src/layouts/PageLayout/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,63 @@
...
+        )}
+      >
+        {title && (
+          <H3>
+            <TextLine>{title}</TextLine>
+          </H3>
+        )}
+        {titleSelect && (
+          <H6 className={cl['bk-page-header__select']}>{titleSelect}</H6>
+        )}
```
mkrause, line 50: I don't think we should wrap the select inside of an `H6`. Haven't tested how voice readers handle this, but I don't think this would translate well.

src/layouts/PageLayout/PageLayout.stories.tsx
mkrause: Ah I see

src/layouts/PageLayout/PageLayout.module.scss
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    .bk-page-layout__header__actions {
+      display: flex;
+      flex-direction: row;
+      gap: bk.$spacing-7;
+    }
+
+    .bk-page-layout__header__h3 {
+      padding-inline-end: bk.$spacing-10;
```
mkrause, line 38: Please use a `gap` for this instead of a padding

src/layouts/PageLayout/PageLayout.module.scss
```diff
@@ -0,0 +1,49 @@
...
+      align-items: center;
+    }
+
+    .bk-page-layout__header__actions {
+      display: flex;
+      flex-direction: row;
+      gap: bk.$spacing-7;
+    }
+
+    .bk-page-layout__header__h3 {
```
mkrause, line 37: Nest the two selectors above inside of `bk-page-layout__header`.

Also, rename `h3` here to "heading" so it's not hardcoding the level.

src/layouts/PageLayout/PageLayout.tsx
```diff
@@ -0,0 +1,90 @@
...
+      {...propsRest}
+      className={cx(
+        'bk',
+        { [cl['bk-page-layout__header']]: !unstyled },
+        propsRest.className,
+      )}
+    >
+      {title && (
+        <H3 className={cl['bk-page-layout__header__h3']}>
+          <TextLine>{title}</TextLine>
```
mkrause, line 50: This `TextLine` doesn't work at the moment since the parent is not a flex container. But also I don't think `TextLine` is the right solution here since I don't think we should have a tooltip on the header in case of overflow.

Instead, you can just use the `bk.text-one-line` mixin on the `<H3>` to get the same effect. Also, add a story with a long heading to show it working.

src/layouts/PageLayout/PageLayout.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+type PageHeaderProps = React.PropsWithChildren<ComponentProps<'header'> & {
+  /** Whether this component should be unstyled. */
+  unstyled?: undefined | boolean,
+  
+  /** A page title to be displayed. */
+  title?: undefined | string,
+  
+  /** A custom scope switcher to use as a title, such as a select. */
+  scopeSwitcher?: undefined | React.ReactNode;
```
mkrause, line 31: Change the semicolon to a comma here

src/layouts/PageLayout/PageLayout.module.scss
```diff
@@ -0,0 +1,49 @@
...
+      padding-inline-end: bk.$spacing-10;
+
+      > span {
+        vertical-align: middle;
+      }
```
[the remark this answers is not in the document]
nighto: Was probably something for the TextLine, but since we removed that, also removed this bit.

PR #373, opened 2025-06-17
Fix tooltip position issues during exit transition (+ some refactoring)

### LINKED ISSUE #263: `TooltipProvider`: tooltips remain on screen (as non-popover) for a little while after they deactivate
There's an issue with tooltips in Firefox during the closing animation. The positioning is off, and can even cause a viewport scrollbar.

### LINKED ISSUE #338: Scroll bar is displayed after tooltip is shown on `DataTable`
When a `Tooltip` appears on table data from mouseover event, a scrollbar is shown on the `DataTable`.
It would be better if the scrollbar doesn't appear.

https://github.com/user-attachments/assets/4362d40f-8d12-4093-9fc2-06512c04ef6c

mkrause:
Resolves #263
Resolves #338

This PR:
- Fixes an issue where tooltips would not be positioned correctly and/or caused scrollbars to appear during the exit transition.
- Fixes tooltips to have a border radius of `bk.$radius-2` (as opposed to the legacy `bk.$radius-s` token).
- Refactoring: move the overlay-related styling from `Tooltip` to `TooltipProvider`.
- Refactoring: `useFloatingElement` now allows `triggerAction` prop to change over time.
- Renames `MenuProvider` and `MenuMultiProvider`'s `action` property to `triggerAction` to be consistent with `TooltipProvider`. The `action` property is still available as an alias, just deprecated.

In Chrome:

https://github.com/user-attachments/assets/b01e4e6d-80f8-4535-8ea7-995e626459e9

In Firefox:

https://github.com/user-attachments/assets/7ffc5220-7f7f-4e21-ac06-b3266363df38

Note that in Chrome there is an exit animation, but in Firefox there isn't. This is because Firefox does not yet support [exit animations](https://developer.chrome.com/blog/entry-exit-animations), in particular the [overlay](https://caniuse.com/mdn-css_properties_overlay) property. Without that support, there is no way to keep the tooltip popover in the top layer while the exit animation runs. This was also why the bug occurred in the first place, since the tooltip was missing the `display: none` in the case that it wasn't in the top layer (`:popover-open` state).


arn-craft (approved):

spli02:
I noticed that full text of some email column is not visible even when screen size is not changed.
Is it better to use `TextLine` for it?
Screenshot from 2025-06-18 09-39-18
As for the issue, it's fixed in my environment.

spli02 (approved):

mkrause:
@spli02 Thanks. Yes, `TextLine` would be better in that story, just hasn't been added yet.

PR #411, opened 2025-07-10
`DialogModal` with in `DialogOverlay` is getting auto closed

### LINKED ISSUE #410: `DialogModal` with in `DialogOverlay` is getting auto closed
When trying to open the `DialogModel` inside of the `DialogOverlay`. The `DialogModel` gets closed after few seconds. Able to reproduce this in story book.

https://github.com/user-attachments/assets/c17eeb24-f7d3-4536-a9ce-203c293eed70

Mahek-Vasoya:
#410 resolved

https://github.com/user-attachments/assets/597ffc8e-d0dc-4a65-b933-31e0cb537384

### DIFF
--- a/src/components/util/overlays/popover/usePopover.ts
+++ b/src/components/util/overlays/popover/usePopover.ts
@@ -30,7 +30,7 @@ export const usePopover = <E extends HTMLElement>(
   controller: PopoverController,
   options: undefined | UsePopoverOptions = {},
 ): PopoverProps<E> => {
-  const { popoverBehavior = 'auto' } = options ?? {};
+  const { popoverBehavior = 'manual' } = options ?? {};
   
   const popoverRef = React.useRef<E>(null);
   
@@ -78,6 +78,35 @@ export const usePopover = <E extends HTMLElement>(
       controller.deactivate();
     }
   }, [controller.activate, controller.deactivate]);
+  
+  // Added for outside click detection
+  React.useEffect(() => {
+  if (!controller.active) return;
+
+  const handleClickOutside = (event: MouseEvent) => {
+    const target = event.target as Node;
+    const popover = popoverRef.current;
+    
+    // Prevent popover from closing if click was inside the popover
+    if (!popover || popover.contains(target)) return;
+
+    // Prevent popover from closing if click was inside portal
+    const isInPortal = (target as HTMLElement).closest('[data-portal]');
+    if (isInPortal) return;
+
+    // Otherwise, clicked outside both → close the popover
+    controller.deactivate();
+  };
+  
+  // Listen for outside clicks
+  document.addEventListener('mousedown', handleClickOutside, true);
+  
+  // Clean up when component unmounts or controller becomes inactive
+  return () => {
+    document.removeEventListener('mousedown', handleClickOutside, true);
+  };
+}, [controller.active, controller.deactivate]);
+
   
   // Sync when the ref changes. This helps prevent time issues where `active` is set to `true`, but the popover is not
   // yet mounted (and thus the ref is `null`). In that case our sync `useEffect` will be too early.
--- a/src/components/util/overlays/popover/usePopover.ts
+++ b/src/components/util/overlays/popover/usePopover.ts
@@ -24,7 +24,7 @@ export type PopoverProps<E extends HTMLElement> = {
 };
 
 /*
- * A utility hook to control the state of a <popover> element used as a modal (with `.showModal()`).
+ * A utility hook to control the state of a popover element (e.g. `<div popover>`).
  */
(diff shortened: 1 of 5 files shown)


src/components/util/overlays/popover/usePopover.ts
```diff
@@ -79,6 +79,35 @@ export const usePopover = <E extends HTMLElement>(
...
+  };
+  
+  // Listen for outside clicks
+  document.addEventListener('mousedown', handleClickOutside, true);
+  
+  // Clean up when component unmounts or controller becomes inactive
+  return () => {
+    document.removeEventListener('mousedown', handleClickOutside, true);
+  };
+}, [controller.active, controller.deactivate]);
```
mkrause, line 108: Please fix the indentation on the useEffect above

src/components/util/overlays/popover/usePopover.ts
```diff
@@ -79,6 +79,35 @@ export const usePopover = <E extends HTMLElement>(
     }
   }, [controller.activate, controller.deactivate]);
   
+  // Added for outside click detection
+  React.useEffect(() => {
+  if (!controller.active) return;
+
+  const handleClickOutside = (event: MouseEvent) => {
+    const target = event.target as Node;
```
mkrause, line 87: Instance of the `as` (which may be unsafe, you can check if `event.target` is actually a node through `event.target instanceof Node`

src/components/util/overlays/popover/usePopover.ts
```diff
@@ -79,6 +79,35 @@ export const usePopover = <E extends HTMLElement>(
...
+    const popover = popoverRef.current;
+    
+    // Prevent popover from closing if click was inside the popover
+    if (!popover || popover.contains(target)) return;
+
+    // Prevent popover from closing if click was inside portal
+    const isInPortal = (target as HTMLElement).closest('[data-portal]');
+    if (isInPortal) return;
```
mkrause, line 95: In the `if` here (and the one above, add braces around the `return;`

src/components/util/overlays/popover/usePopover.ts
```diff
@@ -79,6 +79,35 @@ export const usePopover = <E extends HTMLElement>(
...
+
+  const handleClickOutside = (event: MouseEvent) => {
+    const target = event.target as Node;
+    const popover = popoverRef.current;
+    
+    // Prevent popover from closing if click was inside the popover
+    if (!popover || popover.contains(target)) return;
+
+    // Prevent popover from closing if click was inside portal
+    const isInPortal = (target as HTMLElement).closest('[data-portal]');
```
mkrause, line 94: Here also, check to see if `target` is actually an `HTMLElement` (and then you won't need the `Node` check above since `HTMLElement` is a subtype)

mkrause (changes requested):

src/components/util/overlays/popover/usePopover.ts
```diff
@@ -30,7 +30,7 @@ export const usePopover = <E extends HTMLElement>(
   controller: PopoverController,
   options: undefined | UsePopoverOptions = {},
 ): PopoverProps<E> => {
-  const { popoverBehavior = 'auto' } = options ?? {};
+  const { popoverBehavior = 'manual' } = options ?? {};
```
mkrause, line 33: We shouldn't change the default. This can be overridden from the consumer instead

src/components/util/overlays/popover/usePopover.ts
```diff
@@ -79,6 +79,47 @@ export const usePopover = <E extends HTMLElement>(
     }
   }, [controller.activate, controller.deactivate]);
   
+  // Added for outside click detection
+  React.useEffect(() => {
```
mkrause, line 83: This effect shouldn't run for `popupBehavior` of `auto`.

src/components/util/overlays/popover/usePopover.ts
```diff
@@ -79,6 +79,47 @@ export const usePopover = <E extends HTMLElement>(
...
+    };
+    
+    // Listen for outside clicks
+    document.addEventListener('mousedown', handleClickOutside, true);
+    
+    // Clean up when component unmounts or controller becomes inactive
+    return () => {
+      document.removeEventListener('mousedown', handleClickOutside, true);
+    };
+  }, [controller.active, controller.deactivate]);
```
mkrause, line 119: The logic here has an issue: it will assume that _any_ content inside a portal should not close the popover. But this should only hold for portals that came from within the popover itself.

mkrause (changes requested):

Mahek-Vasoya:
Closing this PR as we have got this resolved with `renderMethod`

PR #423, opened 2025-07-18
`DataTable` – Support table overflow through horizontal scroll

### LINKED ISSUE #414: Support horizontal scrolling in the DataTable when columns overflow

arn-craft:
Resolves #414

### DIFF
--- a/src/components/tables/DataTable/plugins/useRowSelectColumn.tsx
+++ b/src/components/tables/DataTable/plugins/useRowSelectColumn.tsx
@@ -7,37 +7,12 @@ import type * as ReactTable from 'react-table';
 import { Checkbox } from '../../../forms/controls/Checkbox/Checkbox.tsx';
 import cl from './useRowSelectColumn.module.scss';
 
-
 // `react-table` plugin for row selection column. Note: depends on `react-table`'s `useRowSelect` plugin.
 export const useRowSelectColumn = <D extends object>(hooks: ReactTable.Hooks<D>): void => {
-  // Prepend a column with row selection checkboxes
-  hooks.visibleColumns.push(columns => [
-    {
-      id: 'selection',
-      className: cl['bk-data-table-row-select'],
-      Header: ({ getToggleAllPageRowsSelectedProps }) => {
-        const { checked, onChange } = getToggleAllPageRowsSelectedProps();
-        return (
-          <Checkbox aria-label="Select all rows" checked={checked} onChange={onChange}/>
-        );
-      },
-      Cell: ({ row }: ReactTable.CellProps<D, null>) => {
-        const { checked, onChange } = row.getToggleRowSelectedProps();
-        return (
-          <Checkbox aria-label="Select row" checked={checked} onChange={onChange}/>
-        );
-      },
-    },
-    ...columns,
-  ]);
-};
-
-export const useRowSelectColumnMerged = <D extends object>(
-  targetColumnId: string,
-) => (hooks: ReactTable.Hooks<D>): void => {
+  // Merge first column with row selection checkboxes
   hooks.visibleColumns.push(columns =>
-    columns.map((col: ReactTable.ColumnInstance<D>) => {
-      if (col.id === targetColumnId) {
+    columns.map((col: ReactTable.ColumnInstance<D>, index: number) => {
+      if (index === 0) {
         return {
           ...col,
           Header: (headerProps: ReactTable.HeaderProps<D>) => {
--- a/src/components/tables/DataTable/table/DataTable.module.scss
+++ b/src/components/tables/DataTable/table/DataTable.module.scss
@@ -14,7 +14,7 @@
       margin: bk.$spacing-5 0 bk.$spacing-5 0;
     }
     
-    .bk-data-table__scrolled-left {
+    .bk-data-table-container--scrolled-left {
       table.bk-data-table__table {
         thead, tbody {
           th, td {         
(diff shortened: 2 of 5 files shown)


src/components/tables/DataTable/plugins/useRowSelectColumn.tsx
```diff
@@ -5,7 +5,7 @@
 import type * as ReactTable from 'react-table';
 
 import { Checkbox } from '../../../forms/controls/Checkbox/Checkbox.tsx';
-
+import { classNames as cx } from '../../../../util/componentUtil.ts';
```
mkrause, line 8: This import seems to be unused

src/components/tables/DataTable/table/DataTable.module.scss
```diff
@@ -48,6 +100,35 @@
...
+          &.column-sticky--right {
+            right: 0;
+          }
+        }
+
+        th.column-sticky,
+        td.column-sticky {
+          background-color: bk.$theme-card-background-default;
+          // For sticky cells we need a solid colour as background,
+          // otherwise during scroll the content scrolled behind sticky cell will appear as overlapping
```
mkrause, line 117: Let's use this background color for the entire table, not just the sticky column.

src/types/react-table.d.ts
```diff
@@ -60,6 +60,15 @@ interface CustomColumnProps {
...
   label?: string,
+  bkColumnWidth?: {
+    flex: number,
+    width: string,
+  },
+}
+
+interface CustomTableConfig {
+  identifierColumnConfig?: { columnId: string, sticky?: boolean },
+  actionColumnConfig?: { columnId: string, sticky?: boolean },
```
mkrause, line 71: I don't think having a `columnId` config here would be useful. The "identifier" column must always be the first column, otherwise the sticky behavior doesn't make sense. And similarly the "actions" column must be the last column. So I think it would be sufficient for the user to specify (1) whether the first column should be sticky, and (2) whether the last column should be sticky.

src/components/tables/DataTable/table/DataTable.tsx
```diff
@@ -3,6 +3,7 @@
 |* the MPL was not distributed with this file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import type * as React from 'react';
+import { Children, isValidElement, cloneElement, useRef, useState, useEffect } from 'react';
```
mkrause, line 6: Use a `React` namespace for these.

Change `import type * as React from 'react'` to `import * as React from 'react'` and then use `React.` for all these imports.

src/components/tables/DataTable/table/DataTable.tsx
```diff
@@ -18,6 +19,167 @@ import type { DataTableStatus } from '../DataTableContext.tsx';
...
+        const baseStyle = element.props.style ?? {};
+        
+        const style = isFlexColumn(col)
+          ? {
+              ...baseStyle,
+              width: getFlexColumnWidth(col),
+            }
+          : {
+              ...baseStyle,
+              width: '10ch',
```
mkrause, line 93: Change the `width` properties to `inline-size`

src/components/tables/DataTable/table/DataTable.tsx
```diff
@@ -18,6 +19,167 @@ import type { DataTableStatus } from '../DataTableContext.tsx';
...
+  const totalFlex = allColumns.reduce((total, column) => total + getFlexValueFromColumn(column), 0);
+  
+  if (targetColumnFlex === 0) {
+    return 0;
+  }
+  
+  return totalFlex > 0 ? Math.round((targetColumnFlex / totalFlex) * 100) : 0;
+};
+
+export function useCustomColGroup<D extends object>(
```
mkrause, line 60: Why is this a hook? I don't see any hooks in this function. Also please convert it to an arrow function

mkrause (changes requested):

src/components/tables/DataTable/table/DataTable.tsx
```diff
@@ -18,6 +19,115 @@ import type { DataTableStatus } from '../DataTableContext.tsx';
...
+    return 0;
+  }
+  
+  return totalFlex > 0 ? Math.round((targetColumnFlex / totalFlex) * 100) : 0;
+};
+
+export const generateCustomColGroup = <D extends object>(
+  allColumns: ReactTable.ColumnInstance<D>[],
+  columnGroups?: React.ReactNode): React.ReactNode | null => {
+  if (columnGroups && React.isValidElement(columnGroups) && columnGroups.type === 'colgroup') {
```
mkrause, line 59: I think just `if (columnGroups)` here is sufficient, you don't need to check the element type.

mkrause (changes requested):

src/components/tables/DataTable/table/DataTable.tsx
```diff
@@ -40,110 +152,176 @@ export const DataTable = <D extends object>(props: DataTableProps<D>) => {
...
-      <thead>
-        <tr key={key} {...HeaderGroupPropsRest}>
-          {/*<th/> {/ * Empty header for the selection checkbox column */}
+    <>
+      <div
+        {...scrollProps}
+        ref={scrollWrapperRef}
+        className={cx(cl['bk-data-table__scroll-wrapper'], {
+            [cl['bk-data-table__scrolled-left']]: overflowPosition === 'left' || overflowPosition === 'center',
+            [cl['bk-data-table__scrolled-right']]: overflowPosition === 'right' || overflowPosition === 'center',
```
mkrause, line 216: Maybe change CSS classes from using `__` to `--` as per BEM standards

PR #455, opened 2025-08-13
(#401) Add type=search in InputSearch component

### LINKED ISSUE #401: Add `type="search"` in `InputSearch` component to improve accessibility
Currently, `InputSearch` component does not specify an input type. To improve accessibility, it is better to pass `type="search"` to `Input` component.

InputSearch.tsx (L10)
```
export const InputSearch = (props: React.ComponentProps<typeof Input>) =>
  <Input
    icon="search"
    iconLabel="Search"
    placeholder="Search"
    {...props}
  />;
```

Ref: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/search

spli02:
Resolves #401

### DIFF
--- a/src/components/forms/controls/Input/InputSearch.tsx
+++ b/src/components/forms/controls/Input/InputSearch.tsx
@@ -9,6 +9,7 @@ import { Input } from './Input.tsx';
 /** A basic search input control. */
 export const InputSearch = (props: React.ComponentProps<typeof Input>) =>
   <Input
+    type="search"
     icon="search"
     iconLabel="Search"
     placeholder="Search"


mkrause:
So as far as I can tell, adding `type="search"` doesn't have any accessibility implications? For that we need something like `<search>` or `<form role="search">`. What `type="search"` does do is add some extra UI in certain browsers, for example here are Chrome and Safari (respectively) with `type="search"`:

Screenshot 2025-08-14 at 13 39 00

Screenshot 2025-08-14 at 13 39 18

The "x" (which when clicks resets the input) here is added by the browser. It looks okay to me, doesn't break any styling. So I'm okay with the change.

mkrause (approved):

PR #456, opened 2025-08-13
(#402) Add role=switch in Switch component

### LINKED ISSUE #402: Add `role="switch"` on `Switch` component to improve accessibility
The current `Switch` component does not include aria role.
It's better to add `role="switch"` to improve accessibility for screen reader users.

Switch.tsx (L39):

```
    return (
      <Checkbox
        //switch // https://webkit.org/blog/15054/an-html-switch-control
        unstyled
        {...propsRest}
        disabled={!isInteractive}
        className={cx(
          { [cl['bk-switch']]: !unstyled },
          propsRest.className,
        )}
      />
    );
```

Ref: https://www.w3.org/WAI/ARIA/apg/patterns/switch/examples/switch-checkbox/

spli02:
Resolves #402

### DIFF
--- a/src/components/forms/controls/Switch/Switch.tsx
+++ b/src/components/forms/controls/Switch/Switch.tsx
@@ -40,6 +40,7 @@ export const Switch = Object.assign(
       <Checkbox
         //switch // https://webkit.org/blog/15054/an-html-switch-control
         unstyled
+        role="switch"
         {...propsRest}
         disabled={!isInteractive}
         className={cx(


mkrause:
Thanks @spli02. I did some research, found this article on using `<input type="checkbox">` with `role="switch":

https://adrianroselli.com/2021/10/switch-role-support.html

Seems all good as far as I can tell.

mkrause (approved):

PR #471, opened 2025-08-29
(#462) Remove the underline from PageLayout.ScopeSwitcher

### LINKED ISSUE #462: Remove the underline from the `PageLayout.ScopeSwitcher` component
We should remove the underline from the `PageLayout.ScopeSwitcher` component to align with the design system.
(Note: this issue will be addressed by SAF-1623.)

spli02:
Resolves #462 
Screenshot from 2025-08-29 17-43-42

### DIFF
--- a/src/layouts/PageLayout/PageLayout.module.scss
+++ b/src/layouts/PageLayout/PageLayout.module.scss
@@ -40,6 +40,10 @@
 
       .bk-page-layout__header__scope-switcher {
         @include typography.h6;
+        
+        > div {
+          border: none;
+        }
       }
     }
   }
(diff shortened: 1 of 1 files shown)


src/layouts/PageLayout/PageLayout.module.scss
```diff
@@ -40,6 +40,10 @@
 
       .bk-page-layout__header__scope-switcher {
         @include typography.h6;
+        
+        > div {
+          border: none;
+        }
```
mkrause, line 46: Suggestion: you can move this to the `baklava.overrides` layer to ensure it overrides the `Select` component (which is in the same `baklava.components` layer):

```css
@layer baklava.components {
  ...
}
@layer baklava.overrides {
  .bk-page-layout__header__scope-switcher > div {
    border: none;
  }
}
```

mkrause (changes requested):

PR #508, opened 2025-10-02
aria-label related stories

nighto:


src/components/actions/Button/Button.module.scss
```diff
@@ -37,7 +37,10 @@
     
     @include bk.font(bk.$font-family-display, bk.$font-weight-semibold);
     font-size: bk.$font-size-s;
-    text-transform: uppercase;
+
+    .bk-button__all-caps {
```
mkrause, line 41: Let's rename this to `bk-button__content`

PR #530, opened 2025-11-21
Add deduplication check for toast notifications using dedupeKey

midhun-pk:


src/components/overlays/ToastProvider/ToastProvider.stories.tsx
```diff
@@ -138,3 +138,40 @@ export const ToastWithDismiss: Story = {
     ),
   },
 };
+
+export const ToastWithDeduplication: Story = {
+  args: {
+    children: (
+      <>
+        <p>
+          <Button kind="primary" label="Notify (success)"
```
mkrause: Can you clarify (through the button labels maybe) what each button does? Currently when I look at the story there's no context. Also, tip: you can add a doc block (`/**`) above the story export here with some text/Markdown that will be shown in the Storybook doc.

mkrause (changes requested):

PR #542, opened 2025-11-28
Implement error layout on baklava

YounesChellaf:


mkrause:
Declining this in favor of https://github.com/fortanix/baklava/pull/551

PR #556, opened 2025-12-15
Added ListBoxMultiLazy component

midhun-pk:
Added a new ListBoxMultiLazy component that supports loading and selecting multiple items.

The component provides two data-loading modes:

**Scroll**: Behaves similarly to the existing ListBoxLazy component, where additional items are fetched automatically as the user scrolls near the end of the list.

**Custom**: Allows consumers to define their own trigger (e.g., a button click) to load more items. In this mode, props like onUpdateLimit, pageSize, limit, and hasMoreItems are optional, since the consumer fully controls the loading behavior.

ListBoxMultiLazy component also dynamically recalculates the height of each item based on the rendered content, allowing it to fully support rows with variable heights.

**ListBoxMulti.Option** has been updated to forward the ref provided by ListBoxMultiLazy. This ref is required for measuring the rendered element and determining its actual height.

**ListBoxLazy** has also been updated to support the two modes mentioned above.

New stories, **“List Box Lazy With Custom Load More Items Trigger”** and **“List Box Multi Lazy With Custom Load More Items Trigger”**, has been added alongside the existing ListBoxLazy stories to show usage of the custom trigger mode.

### DIFF
--- a/src/components/forms/controls/ListBoxMulti/ListBoxMulti.tsx
+++ b/src/components/forms/controls/ListBoxMulti/ListBoxMulti.tsx
@@ -386,6 +386,7 @@ export const LoadingSpinner = (props: React.ComponentProps<'span'>) => {
         cl['bk-list-box-multi__item'],
         cl['bk-list-box-multi__item--static'],
         cl['bk-list-box-multi__item--loading'],
+        props.className,
       )}
     >
       Loading... <Spinner inline size="small" />
--- a/src/components/forms/controls/ListBoxMultiLazy/ListBoxMultiLazy.tsx
+++ b/src/components/forms/controls/ListBoxMultiLazy/ListBoxMultiLazy.tsx
@@ -26,7 +26,6 @@ import {
   LoadingSpinner,
   ListBoxMultiClassNames,
 } from '../ListBoxMulti/ListBoxMulti.tsx';
-import { Spinner } from '../../../graphics/Spinner/Spinner.tsx';
 
 import cl from './ListBoxMultiLazy.module.scss';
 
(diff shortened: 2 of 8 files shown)


src/components/forms/controls/ListBoxMulti/ListBoxMulti.tsx
```diff
@@ -107,7 +107,7 @@ export type OptionProps = ComponentProps<typeof Button> & {
  * A list box item that can be selected.
  */
 export const Option = (props: OptionProps) => {
-  const { unstyled, itemKey, label, icon, iconDecoration, onSelect, Icon = BkIcon, ...propsRest } = props;
+  const { ref, unstyled, itemKey, label, icon, iconDecoration, onSelect, Icon = BkIcon, ...propsRest } = props;
```
mkrause: Good catch on the missing `ref`, please add it in `ListBox.tsx` also

src/components/forms/controls/ListBoxMultiLazy/ListBoxMultiLazy.tsx
```diff
@@ -0,0 +1,362 @@
...
+  }, [virtualItemKeys]);
+  
+  const isEmpty = useListBoxSelector(state => state.isEmpty());
+
+  const virtualizer = useVirtualizer({
+    count: virtualItemKeys.length,
+    getScrollElement: () => scrollElement,
+    getItemKey,
+    estimateSize: () => 35,
+    measureElement: element => element.getBoundingClientRect().height,
```
mkrause: Shall we add the same to the regular `ListBoxLazy`?

mkrause (changes requested):

PR #557, opened 2025-12-16
`DataTable` select all event bubbling bug fix

Mahek-Vasoya:
Fixed the event bubbling issue for select all `onClick` in the `useRowSelectColumn` plugin function.

Select all `onClick` events that were bubbling up to cause the header click, which was triggering the sort for the first column.

https://fortanix.atlassian.net/browse/KI-4398?atlOrigin=eyJpIjoiMTI5MWVlNzk2MzhmNDY2NTllMzRiNWM0OGI0NjdmY2UiLCJwIjoiaiJ9

### DIFF
--- a/src/components/tables/DataTable/plugins/useRowSelectColumn.tsx
+++ b/src/components/tables/DataTable/plugins/useRowSelectColumn.tsx
@@ -32,11 +32,9 @@ export const useRowSelectColumn = <D extends object>(hooks: ReactTable.Hooks<D>)
                   <Checkbox
                     aria-label="Select all rows"
                     checked={checked}
-                    onClick={(e) => e.stopPropagation()}
-                    onChange={(e) => {
-                      e.stopPropagation();
-                      onChange?.(e);
-                    }}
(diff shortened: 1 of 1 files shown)


src/components/tables/DataTable/plugins/useRowSelectColumn.tsx
```diff
@@ -32,7 +32,11 @@ export const useRowSelectColumn = <D extends object>(hooks: ReactTable.Hooks<D>)
                   <Checkbox
                     aria-label="Select all rows"
                     checked={checked}
-                    onChange={onChange}
+                    onClick={(e) => e.stopPropagation()}
```
mkrause: Some quick code style requests:

- Name the event param `event`, for consistency
- Don't return the result of `stopPropagation` to the `onClick`, wrap the body of the function in `{}`:

```
onClick={event => { event.stopPropagation(); }}
```

mkrause (changes requested):

Issue #485, opened 2025-09-10
Improve interactive element target sizes for accessibility


spli02:
Some interactive elements do not meet the minimum target size requirement defined in WCAG 2.2 (https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html).
To improve accessibility, the target size of the interactive elements should be increased to at least 24×24 CSS pixels.

During my work on the internal app, I found the following issues:

| Component                       | Current Target Size                                         |
|---------------------------------|------------------------------------------------------------|
| `IconButton` (e.g., close button on dialogs, remove button of Tag)                     | 17×17  |
| `Checkbox` of `DataTable`            | 15×15                                                      |
| Sort icon of `DataTable`           | 14×13                                                      |
| Page control button of `DataTable` | 28×17                                                      |
| Pagination button of `DataTable`   | 13×13                                                      |
| `Switch`                          | 40×20                                                      |
| `InputField` / `Select`            | Height: 20px                                               |
| `Radio`                           | 18×18                                                      |

We should also review other components to ensure the minimum target size.

PR #564, opened 2026-01-06
(#485) Improve interactive element target sizes for accessibility

spli02:
Resolves https://github.com/fortanix/baklava/issues/485
(Note: The target size is increased using padding where it does not affect the layout. When padding would impact the layout, the expand-target-area mixin is used instead.)

### DIFF
--- a/src/styling/global/accessibility.scss
+++ b/src/styling/global/accessibility.scss
@@ -110,7 +110,7 @@
 @mixin expand-target-area-relative($offset: vars.$sizing-1) {
   position: relative;
   
-  &:after {
+  &::after {
     content: '';
     position: absolute;
     inset: calc(-1 * $offset);


src/styling/global/accessibility.scss
```diff
@@ -77,6 +77,35 @@
...
+    inset-inline-start: 50%;
+    min-inline-size: $base-width;
+    min-block-size: $base-height;
+    // prevent to make it smaller than the actual element size
+    inline-size: 100%;
+    block-size: 100%;
+    transform: translate(-50%, -50%);
+    cursor: pointer;
+  }
+}
```
mkrause: Shall we add this "relative" variant from alton too?

```scss
// Similar to `expand-target-area` but using relative offset
@mixin expand-target-area-relative($offset: bk.$sizing-1) {
  position: relative;
  
  &:after {
    content: '';
    position: absolute;
    inset: calc(-1 * $offset);
  }
}
```

This allows you to "expand" based on the current width/height.

src/styling/global/accessibility.scss
```diff
@@ -77,6 +77,35 @@
...
+  &::before {
+    content: '';
+    position: absolute;
+    inset-block-start: 50%;
+    inset-inline-start: 50%;
+    min-inline-size: $base-width;
+    min-block-size: $base-height;
+    // prevent to make it smaller than the actual element size
+    inline-size: 100%;
+    block-size: 100%;
```
mkrause: I think we can simplify this to the following:

```scss
    inline-size: max(100%, $base-width);
    block-size: max(100%, $base-height);
```

Or maybe even the following if we want to enforce 24px+:

```scss
    inline-size: max(100%, 24px, $base-width);
    block-size: max(100%, 24px, $base-height);
```

PR #572, opened 2026-01-15
ComboBox components

midhun-pk:
This PR adds the following ComboBox variants and fixes issues in the existing ComboBox component:
1. ComboBoxMulti
2. ComboBoxLazy
3. ComboBoxMultiLazy

### DIFF
--- a/src/components/forms/controls/ListBox/ListBoxStore.tsx
+++ b/src/components/forms/controls/ListBox/ListBoxStore.tsx
@@ -94,9 +94,6 @@ export type ListBoxStateApi = ListBoxState & {
   
   /** Request the given `itemKey` to be selected. If `null`, unset selection. */
   selectItem: (itemKey: null | ItemKey) => void,
-  
-  /** Sync the store state with the DOM. */
-  syncDomState: (options?: { focusVisible?: undefined | boolean }) => void,
 };
 
 // Ref: https://zustand.docs.pmnd.rs/guides/initialize-state-with-props#wrapping-the-context-provider
@@ -159,15 +156,6 @@ export const createListBoxStore = <E extends HTMLElement>(_ref: React.RefObject<
       }
     },
     selectItem: itemKey => { set({ selectedItem: itemKey }); },
-    syncDomState({ focusVisible = true } = {}) {
-      const state = get();
-      const focusedItemKey = state.focusedItem;
-      
-      if (focusedItemKey !== null) {
-        const focusedItemRef = state._internalItemsRegistry.get(focusedItemKey)?.itemRef;
-        focusedItemRef?.current?.focus({ focusVisible });
-      }
-    },
   }));
 };
 export type ListBoxStore = ReturnType<typeof createListBoxStore>;
@@ -234,7 +222,6 @@ export const handleKeyboardInteractions = (store: ListBoxStore) => (event: React
       event.stopPropagation(); // Prevent the key event from triggering other behavior at higher levels
       
       state.focusItem(itemTargetKey);
-      state.syncDomState({ focusVisible: true });
     }
   } catch (error) {
     // If an assumption fails, log the error but don't crash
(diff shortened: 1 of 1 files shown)


src/components/forms/controls/ComboBox/ComboBox.tsx
```diff
@@ -9,17 +9,36 @@ import { mergeProps } from '../../../../util/reactUtil.ts';
...
-export type { ItemKey };
-export type ComboBoxInputProps = ComponentProps<typeof InputDefault>;
+  const handleInternalInputChange = (values: Set<string>) => {
+    setInternalInputValue(values);
+  };
+
+  return {
+    internalInputValue,
+    handleInternalInputChange,
+    isControlled,
```
mkrause: Not sure if we need this `isControlled`. This can be easily derived from the consumer component based on `value`. Also, if we move this to the parent then we can get rid of the `typeof propsRest.value === 'string'` on line 90 and replace it with `!isControlled`.

midhun-pk: Removed

src/components/forms/controls/ComboBox/ComboBox.tsx
```diff
@@ -61,12 +125,50 @@ export const ComboBox = Object.assign(
...
+      if (value === '') {
+        handleSelect(null, null);
+      } else {
+        setInputValue(internalSelected.values().next().value?.label ?? '');
+      }
+    };
+    
+    const selectedFromInternalSelected = React.useMemo(() => {
+      return internalSelected.keys().next().value ?? null;
+    }, [internalSelected]);
```
mkrause: We can remove this (no need to memoize)

src/components/forms/controls/ComboBoxMulti/ComboBoxMulti.stories.tsx
```diff
@@ -0,0 +1,290 @@
...
+};
+
+const ComboBoxMultiInFormC = (props: React.ComponentProps<typeof ComboBoxMulti>) => {
+  return (
+    <>
+      <form
+        id="story-form"
+        onSubmit={event => {
+          event.preventDefault();
+          notify.info(`You have chosen: ${new FormData(event.currentTarget).get('story_component1') || 'none'}`);
```
mkrause: We should change `ComboBoxMulti` (and `ComboBoxMultiLazy`) such that it renders one `<input type="hidden">` per selected value. Then we do `.getAll()` here to get all the values. See `ListBoxMulti.stories.tsx` for an example.

src/components/forms/controls/ComboBoxMulti/ComboBoxMulti.tsx
```diff
@@ -0,0 +1,262 @@
...
+              key={itemKey}
+              className={cl['bk-combo-box__tag']}
+              onRemove={() => { onRemove(new Set([itemKey])); }}
+              content={itemDetails.label}
+            />
+          ))}
+
+          <Button
+            className={cl['bk-combo-box__btn-clear-all']}
+            onPress={() => { onRemove(new Set(selectedOptions.keys())) }}
```
mkrause: Apply the `trimmed` prop to all these "Clear" buttons so that the top/bottom padding doesn't cause a styling issue on tags

Image

src/components/forms/controls/ListBox/ListBoxStore.tsx
```diff
@@ -128,11 +128,6 @@ export const createListBoxStore = <E extends HTMLElement>(_ref: React.RefObject<
       }
     },
     focusItem: itemKey => {
-      const state = get();
-      if (state.focusedItem === itemKey) {
-        // Already focused; avoid redundant focus that might re-enter
-        return;
-      }
```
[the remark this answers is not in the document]
mkrause: UPDATE: I don't see this behavior anymore after testing again, I think my Storybook may have had a cached older version of the code

mkrause (approved):

PR #597, opened 2026-02-24
PropertyList component to match design system requirements

Mahek-Vasoya:
This PR introduces a new implementation of `PropertyList` and `Property` with built-in support for multi-line clamping and expandable values.

### Layout compatibility
- Supports both `horizontal` and `vertical` orientations.
- Allows property sizing via the `size` prop (`small`, `medium`, `large`, `full-width`).
- Uses flexible layout to automatically wrap properties across rows.

### Multi-line clamping (`expandable`)
- Supports configurable `clampLines` (default: `4`).
- Displays ellipsis (`...`) when content overflows.
- Adds a **“View more / View less”** toggle when content exceeds the clamp limit.

### Expand / Collapse behavior
- Expands or collapses the property value without affecting the surrounding layout.
- Toggle is self-contained within the `Property` component.

## Notes
- Clamping is only applied when `value` is a string.
- Layout remains responsive using flex wrapping.

## Example

```tsx
<PropertyList>
  <PropertyList.Property
    size="Large"
    label="Description"
    value={longText}
    expandable
    clampLines={3}
  />
</PropertyList>
```

Screenshot 2026-03-06 at 5 53 44 PM
Screenshot 2026-03-06 at 5 53 54 PM
Screenshot 2026-03-06 at 5 54 01 PM
Screenshot 2026-03-06 at 5 54 09 PM
Screenshot 2026-03-06 at 5 54 13 PM
Screenshot 2026-03-06 at 5 54 19 PM
Screenshot 2026-03-06 at 5 54 25 PM
Screenshot 2026-03-06 at 5 54 29 PM
Screenshot 2026-03-06 at 5 54 36 PM
Screenshot 2026-03-06 at 5 54 40 PM
Screenshot 2026-03-06 at 5 54 45 PM

### DIFF
--- a/src/components/lists/PropertyList/PropertyList.module.scss
+++ b/src/components/lists/PropertyList/PropertyList.module.scss
@@ -65,12 +65,10 @@
       @include bk.text-layout;
     }
 
-    /**
-      * This wrapper is the ONLY element that gets clamped.
-      * React will dynamically set:
-      * style={{ WebkitLineClamp: clampLines }}
-      */
-    .bk-property-list__property__clamped {
+    // This wrapper is the ONLY element that gets clamped.
+    // React will dynamically set:
+    // style={{ WebkitLineClamp: clampLines }}
+    .bk-property-list__property__value--clamped {
       display: -webkit-box;
       -webkit-box-orient: vertical;
       overflow: hidden;
@@ -78,11 +76,12 @@
 
     .bk-property-list__property__toggle-expand {
       align-self: flex-start;
+      cursor: pointer;
+      
       font-weight: bk.$font-weight-regular;
       font-size: bk.$font-size-m;
       color: bk.$theme-button-tertiary-text-default;
-      cursor: pointer;
-      text-decoration: none; // remove default underline
+      text-decoration: none; // Remove default underline
 
       &:hover {
         color: bk.$theme-button-tertiary-text-hover;
--- a/src/components/lists/PropertyList/PropertyList.tsx
+++ b/src/components/lists/PropertyList/PropertyList.tsx
@@ -86,7 +86,7 @@ export const Property = (props: PropertyProps) => {
           className={cx(
             cl['bk-property-list__property__value'],
             {
-              [cl['bk-property-list__property__clamped']]: shouldClamp,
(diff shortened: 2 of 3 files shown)


src/components/lists/PropertyList/PropertyList.module.scss
```diff
@@ -7,19 +7,65 @@
...
     .bk-property-list__property {
-      &.bk-property-list__property--full-width {
-        inline-size: -webkit-fill-available;
+      display: flex;
+      flex-direction: column;
+      gap: bk.$spacing-2;
+
+      /**
+      * Toggle button styling
+      */
```
mkrause: This is using docblock syntax (with the double star, `/**`) but that doesn't have any effect in a CSS file. Can just use a `//` comment.

src/components/lists/PropertyList/PropertyList.tsx
```diff
@@ -13,42 +15,129 @@ export { cl as PropertyListClassNames };
...
+              ? { WebkitLineClamp: clampLines }
+              : undefined
+          }
+        >
+          {value}
+        </div>
+
+        {/* Toggle */}
+        {enableClamping && isStringValue && (
+          <Button
```
mkrause: Maybe you can use `ButtonAsLink` here for the styling. I do see that "View more" doesn't have an underline until you hover, but that's something we could override here I suppose.

src/components/lists/PropertyList/PropertyList.tsx
```diff
@@ -13,42 +15,129 @@ export { cl as PropertyListClassNames };
...
+            shouldClamp
+              ? { WebkitLineClamp: clampLines }
+              : undefined
+          }
+        >
+          {value}
+        </div>
+
+        {/* Toggle */}
+        {enableClamping && isStringValue && (
```
mkrause: I wonder if we also should have a way that we do not show the "View more" button if the text is not actually clamped. This would be hard to do however because we cannot reliably detect if the browser has cut off the text (just counting characters is not sufficient). So currently we leave it up to the consumer to not enable clamping if the text is short?

PR #640, opened 2026-04-10
Fire callback when DialogOverlay is closing via Esc or click outside

### LINKED ISSUE #638: DialogOverlay won't run onRequestClose when closing with native ways (Esc or clicking outside)
Consider the following `DialogOverlay` story:

```
export const DialogOverlayOnRequestClose: Story = {
  args: {
    onRequestClose: () => {
      alert('onRequestClose');
    },
  },  
};
```

It you click on the Cancel button, or the X button on the top right, it would display the alert message instead of closing (as expected). However if you click outside of the `Dialog`, or just press `Esc` to close it, it would not run the function.

Is there a way to make sure it runs that function, other than marking it with `allowUserClose = false` and having to manually reimplement the  closing options?

https://github.com/user-attachments/assets/02a3e116-754c-4c42-a6de-965a6e1cd6d1

nighto:
Gets the job done, but I'm not sure it's the most elegant way. Maybe I am adding a workaround for something that should be fixed elsewhere?

Closes #638 

https://github.com/user-attachments/assets/6e6e8b01-0be8-42fc-92c1-721164a9604b

### DIFF
--- a/src/components/util/overlays/popover/PopoverProvider.tsx
+++ b/src/components/util/overlays/popover/PopoverProvider.tsx
@@ -13,6 +13,7 @@ export type PopoverRef = {
   active: boolean,
   activate: () => void,
   deactivate: () => void,
+  onlyDeactivate: () => void,
 };
 
 export type PopoverParams<E extends HTMLElement> = UsePopoverResult<E> & {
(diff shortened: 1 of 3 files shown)


src/components/util/overlays/popover/PopoverProvider.tsx
```diff
@@ -64,13 +68,16 @@ export const PopoverProvider = Object.assign(
       //source, // TODO: need a reference to the trigger element
       active: active,
       activate: () => { setActiveWithDelay(true); },
-      deactivate: () => { setActiveWithDelay(false); },
+      // we don't want to call props.onRequestClose() when we're closing on our end (X button or Cancel)
+      // otherwise it will fire twice
+      onlyDeactivate: () => { setActiveWithDelay(false); },
```
mkrause, line 73: I don't think we would need this. Apparently without this we're calling `onRequestClose` twice, but then we'd need to figure out why that happens and prevent the double call on `onRequestClose`.

mkrause (changes requested):

mkrause:
Resolved in https://github.com/fortanix/baklava/pull/645

PR #644, opened 2026-04-16
Fixed global filtering behavior when no filterable columns are present on `DataTableEager`

### LINKED ISSUE #643: `DataTableEager` not rendering data on first modal open
When rendering a `DataTableEager` inside a `DialogModal`, the table does not display rows on the first open. However, it renders correctly on subsequent opens.

**Scenario**
- A parent table (`DataTableEager`) is rendered.
- Each row contains an Edit action.
- Clicking Edit opens a `DialogModal`.
- The modal contains another `DataTableEager`.

**Observed Behavior**
- On the first modal open: The inner DataTableEager renders without rows (empty state).
- On subsequent opens: The table renders correctly with data.

**Expected Behavior**
- The inner `DataTableEager` should render with the correct data on the first modal open.

**Key Observation**
- The issue does not occur in React Strict Mode.
- In Strict Mode, the table renders correctly even on the first open.

Mahek-Vasoya:
This PR contains a fix for #643 

This change ensures that `customGlobalFilter` is only enabled when at least one column explicitly allows it.

Previously, global filtering was always active, even when all columns had `disableGlobalFilter: true`  or `disableGlobalFilter: undefined`. This caused the filtering pipeline to run without any eligible columns, resulting in all rows being filtered out and the table appearing empty.

**Changes**
- Added `customGlobalFilter`:

**Before**
- Global filter always applied
- No filterable columns → table rendered empty

**After**
- Global filter runs only when valid columns exist
- No filterable columns → data renders normally

### DIFF
--- a/src/components/tables/DataTable/DataTableEager.tsx
+++ b/src/components/tables/DataTable/DataTableEager.tsx
@@ -18,6 +18,49 @@ export * from './DataTableContext.tsx';
 export { Pagination };
 export { DataTablePlaceholderEmpty, DataTablePlaceholderError } from './table/DataTablePlaceholder.tsx';
 
+/**
+ * Custom global filter for react-table.
+ *
+ * Handles:
+ * - Empty search (returns all rows)
+ * - No filterable columns (returns all rows)
+ * - Safe comparison for null/undefined/non-string values
+ * - Case-insensitive partial matching across all columns
+ */
+const customGlobalFilter: ReactTable.FilterType<D> = (
+  rows: ReactTable.Row<D>[], // All rows before filtering
+  columnIds: string[], // IDs of columns allowed for global filtering
+  filterValue: unknown // Value from global search input
+) => {
+  // Normalize the search input:
+  // - Convert to string
+  // - Handle null/undefined safely
+  // - Make it case-insensitive
+  const search = String(filterValue ?? '').toLowerCase();
+
+  // If search is empty → do not filter, return all rows
+  // Prevents "empty table on initial load" issue
+  if (!search) return rows;
+
+  // If no columns are filterable → skip filtering
+  // Prevents react-table from returning empty results
+  // when all columns have `disableGlobalFilter: true`
+  if (!columnIds.length) return rows;
+
+  // Filter rows:
+  // Include a row if ANY filterable column matches the search value
+  return rows.filter(row =>
+    columnIds.some(id => {
+      const value = row.values[id]; // Get the cell value for the column
+      // Convert value safely to string and compare
+      // - Handles null/undefined
+      // - Supports numbers, booleans, etc.
+      return String(value ?? '')
+        .toLowerCase()
+        .includes(search); // Partial match
+    })
+  );
+};
 
 interface ReactTableOptions<D extends object> extends ReactTable.TableOptions<D> {
   // Add custom properties here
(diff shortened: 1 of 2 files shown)


src/components/tables/DataTable/DataTableEager.tsx
```diff
@@ -82,7 +82,7 @@ export const TableProviderEager = <D extends object>(props: TableProviderEagerPr
       },
       
       // useGlobalFilter
-      manualGlobalFilter: false,
+      manualGlobalFilter: true,
```
mkrause: This now breaks the global filter functionality. See the following story: DataTableEager → Single Page (http://localhost:6006/?path=/story/components-tables-datatable-datatableeager--single-page). Before, the search box works, but in this PR it doesn't filter at all.

mkrause (changes requested):

src/components/tables/DataTable/DataTableEager.tsx
```diff
@@ -18,6 +18,51 @@ export * from './DataTableContext.tsx';
...
+    // Filter rows:
+    // Include a row if ANY filterable column matches the search value
+    return rows.filter(row =>
+      columnIds.some(id => {
+        const value = row.values[id]; // Get the cell value for the column
+        // Convert value safely to string and compare
+        // - Handles null/undefined
+        // - Supports numbers, booleans, etc.
+        return String(value ?? '')
+          .toLowerCase()
```
mkrause, line 65: Suggestion: use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLocaleLowerCase here (as well as above on line 40)

src/components/tables/DataTable/DataTableEager.tsx
```diff
@@ -18,6 +18,51 @@ export * from './DataTableContext.tsx';
...
+    const search = String(filterValue ?? '').toLowerCase();
+
+    // If search is empty → do not filter, return all rows
+    // Prevents "empty table on initial load" issue
+    if (!search) return rows;
+
+    // If no columns are filterable → skip filtering
+    // Prevents react-table from returning empty results
+    // when all columns have `disableGlobalFilter: true`
+    if (!columnIds.length) return rows;
```
mkrause: Can you add the braces around the `if` statement bodies here? I'd prefer not to use single-line `if` syntax (makes it easy to introduce bugs in the future if someone adds a new statement)

src/components/tables/DataTable/DataTableEager.tsx
```diff
@@ -18,6 +18,51 @@ export * from './DataTableContext.tsx';
...
+ * - Empty search (returns all rows)
+ * - No filterable columns (returns all rows)
+ * - Safe comparison for null/undefined/non-string values
+ * - Case-insensitive partial matching across all columns
+ */
+const customGlobalFilter = <D extends object>(): ReactTable.FilterType<D> => {
+  return (
+    rows: ReactTable.Row<D>[], // All rows before filtering
+    columnIds: string[], // IDs of columns allowed for global filtering
+    filterValue: ReactTable.Row<D>[], // Value from global search input
```
mkrause: For the array types here can you use `Array<...>` syntax?

mkrause (changes requested):

src/components/tables/DataTable/DataTableEager.tsx
Mahek-Vasoya: This wouldn’t be the right use case for `toLocaleLowerCase` since it applies locale-specific transformations based on the user/system settings, which can lead to inconsistent matching.

In our case, we want predictable, locale-independent matching for search, so `toLowerCase` is the safer and more appropriate choice.

eg `"I".toLocaleLowerCase('tr')=== "ı" // not "i"`

mkrause (approved):

PR #654, opened 2026-04-29
Highlight table row support

Mahek-Vasoya:


src/components/tables/DataTable/plugins/useRowSelectColumn.tsx
```diff
@@ -162,3 +163,38 @@ export const useRowSelectColumnRadio = <D extends object>(hooks: ReactTable.Hook
...
+      return {
+        ...props, // preserve existing props added by react-table or other plugins
+
+        // Merge existing className with our highlight class
+        className: cx(
+          props.className,
+
+          // Only apply highlight class if this row matches the target id
+          // Using `&&` avoids "undefined" class issues
+          row.id === highlightedRowId && cl['bk-data-table-row-highlighted']
```
nighto: minor nit: missing final comma

src/components/tables/DataTable/DataTableEager.stories.tsx
```diff
@@ -586,12 +587,15 @@ const useTableModal = () => {
...
     },
-    render() {
+    render({ onClose }: { onClose: () => void }) {
       return (
         <DialogModal
           modalRef={modal}
           title="Modal Edit"
           size="large"
+          onClose={() => {
+            onClose?.()
```
nighto: semicolon

src/components/tables/DataTable/DataTableEager.stories.tsx
```diff
@@ -609,38 +613,136 @@ const useTableModal = () => {
...
+    render({ onClose }: { onClose?: () => void }) {
+      return (
+        <DialogOverlay
+          popoverRef={overlay}
+          title="Overlay Edit"
+          display="slide-over"
+          size="medium"
+          onToggle={(event) => {
+            if (event.newState === 'closed') {
+              onClose?.()
```
nighto: here too

src/components/tables/DataTable/DataTableEager.stories.tsx
```diff
@@ -609,38 +613,136 @@ const useTableModal = () => {
...
+            <IconButton
+              icon="edit"
+              label="edit"
+              onPress={() => {
+                dataTableModal.activate();
+                setHighlightedRow(row.id);
+              }}
+            />
+          );
+        }
```
nighto: comma

src/components/tables/DataTable/DataTableEager.stories.tsx
```diff
@@ -609,38 +613,136 @@ const useTableModal = () => {
...
               icon="edit"
               label="edit"
               onPress={() => {
                 dataTableModal.activate();
+                setHighlightedRow(row.id);
               }}
             />
-          </>)
+          );
+        }
```
nighto: comma

nighto (approved):
LGTM 🚀

imadbakir:
Thanks Mahek! Conceptually I think it's fine, but the whole plugin thing seems to me like an unnecessary layer of abstraction, for something that is rather simple and doesn't require access to react-table API.
I think a simple conditional class directly on the table row would suffice perfectly and reduce the amount of code.
We can expose this as a simple prop on the DataTable itself, something like `highlightedRowId?: string`, and then pass it down to where the row is rendered.
In `DataTable.tsx`, around line 279, it would look something like this:

```tsx
<tr
  {...rowProps}
  className={cx(
    rowProps.className,
    ...
    { [cl['bk-data-table__row--highlighted']]: row.id === highlightedRowId },
    ...
  )}
>```

Then the styling can also live next to the other row styles in DataTable.module.scss, instead of being part of the row selection plugin.

File: src/components/tables/DataTable/DataTableEager.stories.tsx [hunk 6]
```diff
@@ -609,38 +613,136 @@ const useTableModal = () => {
...
+            if (highlightedRow) {
+              setHighlightedRow(null);
+            }
+          },
+        })}
+        <DataTableEager.TableProviderEager
+          columns={tableCol}
+          items={data}
+          getRowId={(row: User) => row.id}
+          plugins={[DataTablePlugins.useRowSelectColumn, DataTablePlugins.useRowHighlight(highlightedRow)]}
```
mkrause: Using it as a function argument here means we're reconstructing the plugin on every render. Not sure if that's a good idea. May be better to pass in the `highlightedRow` separately?

imadbakir (approved):

PR #664, opened 2026-05-14
Combobox Improvements

midhun-pk:
The current `ComboBox` implementation is tightly coupled to selection-based use cases and is not designed as a generic popover component. As a result, it does not support more complex or dynamic popover content such as DatePicker, MultiSearch, and similar components.

During discussions, it was concluded that components requiring custom popover behavior should be built directly using existing utilities such as `MenuProvider`, `useFloatingElement`, and related utilities, rather than extending or depending on the `ComboBox` component.

In this PR:
- Adds a new trigger-action: `combobox`
- The new interaction opens the popover only on:
       - onClick
       - onKeyDown (Up arrow / Down arrow)
       - input value changes
- The popover closes on focus out
- Rename all existing `ComboBox` components and related variants by prefixing them with Select.

While integrating with DSM, I fount a few issues related to popover open/close behaviour.

This is the updated behaviour:
1. If the input is cleared and then loses focus, the current selection should be cleared.
2. If the input is partially modified but no new option is selected, the previously selected value should be restored in the input


src/components/util/overlays/floating-ui/useFloatingElement.tsx
```diff
@@ -225,8 +393,11 @@ export type UseFloatingElementOptions = {
    * - `focus-interactive`: The floating element will be open when the reference element is focused, or when the user
    *   focuses an element inside of the floating element. Clicking inside the floating element (thus losing focus) will
    *   also not close it, light dismiss will only occur when clicking outside of the floating/reference element.
+   * - `combobox`: The floating element will be open when the reference element is clicked, or when arrow keys are
+   *   pressed. Clicking inside the floating element (thus losing focus) will not close it, light dismiss will 
+   *   only occur when clicking outside of the floating/reference element.
    */
-  triggerAction?: undefined | 'none' | 'click' | 'hover' | 'focus' | 'focus-interactive',
+  triggerAction?: undefined | 'none' | 'click' | 'hover' | 'focus' | 'focus-interactive' | 'combobox',
```
mkrause, line 401: Can you add one or more stories for this new trigger type?

src/components/util/overlays/floating-ui/useFloatingElement.tsx
```diff
@@ -200,6 +200,174 @@ const useFocusInteractive = (context: FloatingContext, options: UseFocusInteract
...
+  const handleReferenceKeyDown = React.useCallback(
+    (event: React.KeyboardEvent) => {
+      if (!enabled) return;
+
+      if (event.key === 'ArrowDown' || event.key === 'ArrowUp') {
+        onOpenChange(true, event.nativeEvent, 'list-navigation');
+      }
+    },
+    [enabled, onOpenChange]
+  );
```
mkrause, line 259: We should prevent the default scroll behavior on arrow keys. Currently, the floating element opens/closes but it also scrolls the page along with it.

mkrause (changes requested):

src/components/util/overlays/floating-ui/useFloatingElement.tsx
```diff
@@ -200,6 +200,174 @@ const useFocusInteractive = (context: FloatingContext, options: UseFocusInteract
...
+  const handleReferenceClick = React.useCallback(
+    (event: React.MouseEvent) => {
+      if (!enabled) return;
+      onOpenChange(true, event.nativeEvent, 'click');
+    },
+    [enabled, onOpenChange]
+  );
+
+  // Arrow keys → open menu
+  const handleReferenceKeyDown = React.useCallback(
```
mkrause, line 249: "Escape" key should also close the floating element.

Or, slightly more generally, any device-dependent "close" event should close it, see `CloseWatcher` which supports "Escape" on desktop, "Back" button on Android, etc.: https://developer.mozilla.org/en-US/docs/Web/API/CloseWatcher

`CloseWatcher` is not supported on Safari yet though, so maybe hardcoding Escape could also work for now (or use a ponyfill?).

mkrause (changes requested):

src/components/util/overlays/floating-ui/useFloatingElement.tsx
```diff
@@ -225,8 +408,11 @@ export type UseFloatingElementOptions = {
    * - `focus-interactive`: The floating element will be open when the reference element is focused, or when the user
    *   focuses an element inside of the floating element. Clicking inside the floating element (thus losing focus) will
    *   also not close it, light dismiss will only occur when clicking outside of the floating/reference element.
+   * - `combobox`: The floating element will be open when the reference element is clicked, or when arrow keys are
+   *   pressed. Clicking inside the floating element (thus losing focus) will not close it, light dismiss will 
+   *   only occur when clicking outside of the floating/reference element.
```
mkrause: Now that the CloseWatcher behavior is implemented, please update the comment here to mention that light dismiss will happen on clicking outside OR when a browser close event happens (e.g. Escape).

src/types/globals.d.ts
```diff
@@ -32,3 +32,24 @@ type TogglePopoverOptions = {
...
+  oncancel: ((this: CloseWatcher, e: CustomEvent) => void) | null;
+}
+
+interface CloseWatcherOptions {
+  signal?: AbortSignal;
+}
+
+declare var CloseWatcher: {
+  prototype: CloseWatcher;
+  new (options?: CloseWatcherOptions): CloseWatcher;
```
mkrause: Can you add the explicit `undefined |` to the `?:` properties in these changes?

src/components/util/overlays/floating-ui/useFloatingElement.tsx
```diff
@@ -200,6 +201,188 @@ const useFocusInteractive = (context: FloatingContext, options: UseFocusInteract
...
 };
 
+type UseComboBoxOptions = {
+  /** Whether to enable this interaction. Useful for conditional usage of the hook. */
+  enabled?: undefined | boolean,
+};
+
+const useComboBoxInteraction = (
+  context: FloatingContext,
+  options: UseComboBoxOptions = {}
```
mkrause: Nit: add the trailing comma here

src/components/util/overlays/floating-ui/useFloatingElement.tsx
```diff
@@ -200,6 +201,188 @@ const useFocusInteractive = (context: FloatingContext, options: UseFocusInteract
...
+): ElementProps => {
+  const { elements, onOpenChange } = context;
+
+  const enabled = options.enabled ?? true;
+
+  const anchorEl = elements.reference;
+  const popoverEl = elements.floating;
+
+  React.useEffect(() => {
+    if (!context.open) return;
```
mkrause: Nit: add braces aroudn `return;` (no single-line `if` statements)

src/util/CloseWatcher.ts
```diff
@@ -0,0 +1,92 @@
...
+
+      this.watcher.onclose = event => {
+        this.destroyed = true;
+        this.handlers.close?.(event as CustomEvent);
+      };
+
+      return;
+    }
+
+    // Safari fallback
```
mkrause: Instead of mentioning "Safari" let's call it "non-supporting browsers". For instance, Firefox has only supported it for a few major versions at this point.

src/components/util/overlays/floating-ui/useFloatingElement.tsx
```diff
@@ -200,6 +201,188 @@ const useFocusInteractive = (context: FloatingContext, options: UseFocusInteract
...
+    };
+  }, [context.open, onOpenChange]);
+
+  // Focus should NOT open menu anymore
+  const handleReferenceFocus = React.useCallback(() => {
+    if (!enabled) return;
+    // no-op
+  }, [enabled]);
+
+  // Click → open menu
```
mkrause: Let's not talk about "menu" here since we don't know what's inside the floating element. Can just call it the "floating element"

src/components/util/overlays/floating-ui/useFloatingElement.tsx
```diff
@@ -200,6 +201,188 @@ const useFocusInteractive = (context: FloatingContext, options: UseFocusInteract
...
+
+  // On change event -> open menu
+  const handleReferenceChange = React.useCallback(
+    (event: React.ChangeEvent<HTMLInputElement>) => {
+      if (!enabled) return;
+
+      onOpenChange(true, event.nativeEvent, 'reference-press');
+    },
+    [enabled, onOpenChange]
+  );
```
mkrause: I'm not sure if this is desirable. Why should the floating element open upon `change` event? If the reference element is an input and I type in it, does that the floating element should open? I would think it shouldn't.

src/components/util/overlays/floating-ui/useFloatingElement.tsx
```diff
@@ -200,6 +201,188 @@ const useFocusInteractive = (context: FloatingContext, options: UseFocusInteract
...
+      const isInside =
+        event.relatedTarget instanceof Node &&
+        ((popoverEl && popoverEl.isConnected && popoverEl.contains(event.relatedTarget)) ||
+          anchorEl.contains(event.relatedTarget));
+
+      if (!isInside) {
+        onOpenChange(false, event.nativeEvent, 'focus-out');
+      } else {
+        event.preventDefault();
+      }
```
mkrause, line 274: Can you explain why this `preventDefault()` is here? It is not there in `useFocusInteractive`.

src/components/util/overlays/floating-ui/useFloatingElement.stories.tsx
```diff
@@ -229,3 +229,91 @@ export const FloatingElementWithSVGAnchor: Story = {
 export const FloatingElementWithPopoverAuto: Story = {
   args: { renderAnchor: buttonAnchor('Click me'), options: { popoverBehavior: 'auto' } },
 };
+
+export const FloatingElementWithTriggerCombobox: Story = {
```
mkrause: Can you move this story up so it's underneath the story for focus interactive trigger?

src/components/util/overlays/floating-ui/useFloatingElement.stories.tsx
```diff
@@ -229,3 +229,91 @@ export const FloatingElementWithSVGAnchor: Story = {
...
+      <p><Button label="Toggle popover" onPointerDown={onTogglePopover}/></p>
+    </>
+  );
+};
+
+export const FloatingElementWithManualTrigger: Story = {
+  render: (args) => <FloatingElementWithManualTriggerC {...args}/>,
+  args: {
+    renderAnchor: buttonAnchor('Focus me'),
+    popoverContent: (
```
mkrause, line 350: The styling is broken in this story, it no longer gets the styles from `FloatingElementControlled`.

src/components/util/overlays/floating-ui/useFloatingElement.stories.tsx
```diff
@@ -229,3 +229,91 @@ export const FloatingElementWithSVGAnchor: Story = {
...
+      {props.renderAnchor(referenceProps)}
+      {isMounted &&
+        <div {...floatingProps}>{props.popoverContent ?? 'This is a popover'}</div>
+      }
+      <p><Button label="Toggle popover" onPointerDown={onTogglePopover}/></p>
+    </>
+  );
+};
+
+export const FloatingElementWithManualTrigger: Story = {
```
mkrause, line 346: Can you add a docblock comment here explaining what this story is testing?

src/components/overlays/MenuMultiProvider/MenuMultiProvider.tsx
```diff
@@ -206,6 +206,7 @@ export const useMenuAnchor = <RenderArgs extends BaseAnchorRenderArgs>(props: Us
         'aria-expanded': isOpen,
         // biome-ignore lint/suspicious/noExplicitAny: `onKeyDown` should be a function here
         onKeyDown: mergeCallbacks([props.onKeyDown as any, onKeyDown]),
+        onBlur: props.onBlur,
```
mkrause: I don't think this line is needed, since there is a `...props` spread above and nothing is overriding `onBlur`.

src/components/overlays/MenuMultiProvider/MenuMultiProvider.tsx
```diff
@@ -367,7 +368,14 @@ export const useMenuSelect = (options: UseMenuSelectHandlerOptions) => {
       const previous = previousActiveElementRef.current;
 
       if (previous) {
-        previous.focus({ focusVisible: false });
+        const el = previous as HTMLElement;
```
mkrause: This seems like an unsafe type cast. Also, I'm not sure why this line is necessary since I believe `previous` is already inferred to be an `HTMLElement` by TypeScript.

mkrause (changes requested):

src/components/util/overlays/floating-ui/useFloatingElement.tsx
midhun-pk: Updated

src/types/globals.d.ts
midhun-pk: Updated

src/components/util/overlays/floating-ui/useFloatingElement.tsx
midhun-pk: Updated

src/components/util/overlays/floating-ui/useFloatingElement.stories.tsx
midhun-pk: Updated

src/components/util/overlays/floating-ui/useFloatingElement.stories.tsx
midhun-pk: Updated

src/components/overlays/MenuMultiProvider/MenuMultiProvider.tsx
midhun-pk: Updated

src/components/util/overlays/floating-ui/useFloatingElement.tsx
midhun-pk: Updated

src/util/CloseWatcher.ts
midhun-pk: Updated

src/components/forms/controls/SelectComboBox/SelectComboBox.stories.tsx
```diff
@@ -46,18 +46,16 @@ export default {
...
+          <SelectComboBox.Option key={fruitKey} itemKey={fruitKey} label={fruitName}/>
         )}
       </>
     ),
   },
-  render: (args) => <ComboBox {...args}/>,
+  render: (args) => <SelectComboBox {...args}/>,
 } satisfies Meta<ComboBoxArgs>;
 
 export const ComboBoxStandard: Story = {};
```
mkrause: Can you update the story names to match the new component name?

src/components/forms/controls/SelectComboBox/SelectComboBox.tsx
```diff
@@ -119,90 +136,143 @@ export type ComboBoxProps = Omit<InputProps, 'onSelect'> & {
...
+      onDropdownBlur?.(evt);
     };
 
     return (
       <MenuProvider
         label={label}
         items={options}
-        role="combobox"
-        triggerAction="focus-interactive" // Keep the dropdown menu open while the input is focused
+        role="listbox"
```
mkrause: The `role` here in `MenuProvider` is passed to `useFloatingMenu`, I guess this should still be `combobox`? The inner `ListBox` itself would still have `role="listbox"`.

src/components/forms/controls/SelectComboBoxLazy/SelectComboBoxLazy.stories.tsx
```diff
@@ -144,36 +144,19 @@ export const ComboBoxLazyInfinte: Story = {
   },
```
mkrause, line 144: There's a typo in this story (already existing): `Infinte` → `Infinite`

src/components/forms/controls/SelectComboBoxMulti/SelectComboBoxMulti.stories.tsx
```diff
@@ -198,32 +205,49 @@ export const ComboBoxMultiUncontrolled: Story = {
   },
 };
 
-const ComboBoxMultiwithFilterC = (props: Partial<React.ComponentProps<typeof ComboBoxMulti>>) => {
+const ComboBoxMultiwithFilterC = (props: Partial<React.ComponentProps<typeof SelectComboBoxMulti>>) => {
```
mkrause: Typo: `ComboBoxMultiwithFilterC` → `ComboBoxMultiWithFilterC`

src/components/forms/controls/SelectComboBoxMulti/SelectComboBoxMulti.tsx
```diff
@@ -114,17 +124,15 @@ const ComboBoxMultiInput = (props: ComboBoxMultiInputProps) => {
     <>
       <Input
         role="combobox"
-        automaticResize
+        automaticResize={automaticResize}
```
mkrause, line 127: This change means `automaticResize` is now false by default rather than true. Should we keep it as is? Browser support for `fiels-sizing` is becoming baseline soon. And for combobox in particular it might be nice for the input to adapt to the content.

(Applies to other components also)

src/components/forms/controls/SelectComboBoxMulti/SelectComboBoxMulti.tsx
```diff
@@ -198,70 +214,118 @@ export type ComboBoxMultiProps = Omit<InputProps, 'onSelect'> & {
...
+      onDropdownBlur?.(evt);
     };
 
     return (
       <MenuMultiProvider
         label={label}
         items={options}
-        role="combobox"
-        triggerAction="focus-interactive" // Keep the dropdown menu open while the input is focused
+        role="listbox"
```
mkrause: Same comment as earlier

mkrause (changes requested):

mkrause (approved):

Issue #672, opened 2026-05-22
Add loading state for `Card` and `Panel`


mkrause:
In `Dialog` we already have something similar:

```
  /** Controls the modal content state to show a loader when in the loading state */
  state?: undefined | 'loading' | 'ready',
```

We should add the same to the `Card` and `Panel` components.

nighto:
(Also update JSDoc coment on Dialog.tsx to read Dialog instead of Panel

PR #673, opened 2026-05-28
Adds loading indicator to Card and Panel components

nighto:


Mahek-Vasoya (approved):

src/components/containers/Card/Card.tsx
```diff
@@ -58,13 +59,19 @@ export type CardProps = ComponentProps<'section'> & {
   
   /** If enabled, removes the outside border/padding. For use in nested contexts. Default: false. */
   flat?: undefined | boolean,
+  
+  /** Whether the panel display a loading indicator, instead of it's children */
+  loading?: undefined | boolean,
```
mkrause: Let's convert this into a `state` enum, so that we can extend in the future. This is also how it works in `Dialog` currently. See the description here: https://github.com/fortanix/baklava/issues/672

src/components/containers/Card/Card.tsx
```diff
@@ -73,10 +80,11 @@ export const Card = Object.assign(
           'bk',
           { [cl['bk-card']]: !unstyled },
           { [cl['bk-card--flat']]: flat },
+          { [cl['bk-card--loading']]: loading },
           propsRest.className,
         )}
       >
-        {children}
+        {loading ? <Spinner size={loadingSize}/> : children}
```
mkrause: When in loading state, we should still show the `children` in the background, but inert.

There will be two cases (for which I expect at least two stories):
1. Content is not empty, then we should display the content and render the spinner as an overlay over the content, with a semi-transparent backdrop and with a fixed position so it stays in the center when the user scrolls the background content.
2. Content is empty (`children` is none). In this case, we can show a spinner as the content. (There should be no semi-transparent backdrop.)

In the empty content case, you may need to add a minimum height so that the spinner is shown. In case (1) where the user has defined content, don't set a min-height because we don't want to mess with the dimensions set by the user.

src/components/containers/Card/Card.tsx
```diff
@@ -58,13 +59,19 @@ export type CardProps = ComponentProps<'section'> & {
   
   /** If enabled, removes the outside border/padding. For use in nested contexts. Default: false. */
   flat?: undefined | boolean,
+  
+  /** Whether the panel display a loading indicator, instead of it's children */
+  loading?: undefined | boolean,
+  
+  /** Spinner loading size, defaults to large. */
+  loadingSize?: undefined | 'small' | 'medium' | 'large',
```
mkrause: Let's remove this prop, and have the component be in control of the loading state + sizing. In the future we may want to change the spinner to something else (e.g. skeleton view) without it affecting the public interface of the component.

I would suggest:
- In your loading state story (with content+scroll), also allow horizontal resizing. You can do this with the decorator `resize` prop:

```
decorators: [Story => <LayoutDecorator size="small" resize="inline"><Story/></LayoutDecorator>],
```

- Then, in your CSS, use a container query to get the current size of the `Card`, and adjust the size of the spinner accordingly.

src/components/containers/Card/Card.stories.tsx
```diff
@@ -171,3 +171,23 @@ export const CardNested: Story = {
...
+    loadingSize: 'medium',
+  },
+};
+
+export const CardLoadingSmall: Story = {
+  args: {
+    loading: true,
+    loadingSize: 'small',
+  },
+};
```
mkrause, line 206: - Remove the size variants (since we're removing the `loadingSize` prop).
- Have three stories here: one without children, one with but without scroll, and one with scroll.

mkrause (changes requested):

src/components/containers/Card/Card.module.scss
```diff
@@ -67,6 +68,47 @@
 @layer baklava.components {
   .bk-card {
     @include bk.component-base(bk-card);
+    // Do not set contain: layout (included in bk.component-base mixin)
+    // because it breaks the variable --card used for anchor positioning
     @include bk-card;
+    
+    contain: style;
+    anchor-name: --card;
+    anchor-scope: --card;
```
mkrause: Could you rename this `--bk-card`? To prevent naming conflicts

src/components/containers/Card/Card.module.scss
```diff
@@ -67,6 +68,47 @@
 @layer baklava.components {
   .bk-card {
     @include bk.component-base(bk-card);
+    // Do not set contain: layout (included in bk.component-base mixin)
+    // because it breaks the variable --card used for anchor positioning
```
mkrause: This comment should be moved down to be above the `contain: style`.

src/components/containers/Card/Card.module.scss
```diff
@@ -67,6 +68,47 @@
...
     @include bk-card;
+    
+    contain: style;
+    anchor-name: --card;
+    anchor-scope: --card;
+    
+    .bk-card__loading {
```
mkrause, line 79: This needs a `z-index: 1`, otherwise it can be rendered behind the content, like here:

Image

src/components/containers/Card/Card.module.scss
```diff
@@ -67,6 +68,47 @@
...
+      place-self: stretch;
+
+      display: grid;
+      place-content: center;
+      
+      background: color-mix(in oklch, black, 80% transparent);
+      
+      .bk-card__loading__spinner {
+        @include spinner.spinner-medium;
+        inline-size: min(20cqmin, 5rem);
```
mkrause: Let's increase it a bit from `20cqmin` to `35cqmin`, I think currently it's a bit small.

src/components/containers/Card/Card.stories.tsx
```diff
@@ -171,3 +177,37 @@ export const CardNested: Story = {
...
+    status: 'loading',
+  },
+};
+
+export const CardLoadingBig: Story = {
+  args: {
+    status: 'loading',
+    children: <><LoremIpsum/><LoremIpsum/><LoremIpsum/><LoremIpsum/><LoremIpsum/></>
+  }
+}
```
mkrause: Nit: semicolon

src/components/containers/Card/Card.stories.tsx
```diff
@@ -171,3 +177,37 @@ export const CardNested: Story = {
...
+export const CardLoading: Story = {
+  args: {
+    status: 'loading',
+  },
+};
+
+export const CardLoadingBig: Story = {
+  args: {
+    status: 'loading',
+    children: <><LoremIpsum/><LoremIpsum/><LoremIpsum/><LoremIpsum/><LoremIpsum/></>
```
mkrause: `LoremIpsum` has a `paragraphs` prop that you can use to render multiple paragraphs (instead of repeating the element)

src/components/containers/Card/Card.tsx
```diff
@@ -73,9 +77,16 @@ export const Card = Object.assign(
...
+          { [cl['bk-card--empty']]: !children },
           propsRest.className,
         )}
+        inert={status === 'loading'}
       >
+        {status === 'loading' && (
+          <div className={cl['bk-card__loading']}>
+            <Spinner className={cl['bk-card__loading__spinner']}/>
+          </div>
+        )}
```
mkrause, line 90: Let's move this down to below the `children` so it doesn't display below the content (the `z-index` should solve the same issue, but can't hurt to do both).

src/components/containers/Panel/Panel.module.scss
```diff
@@ -28,13 +29,51 @@
...
+    
+    // Use a small spinner when there's no content
+    // TODO: Fix it when there's content, but it's tiny (i.e. one line of text)
+    &.bk-panel--empty {
+      .bk-panel__loading {
+        .bk-panel__loading__spinner {
+          @include spinner.spinner-small;
+        }
+      }
+    }
```
mkrause, line 77: All the same comments here

mkrause (changes requested):

mkrause (approved):

PR #677, opened 2026-06-10
Avoid calling `timeout` when `asyncTimeout` is `Infinity` in `Button` and `SubmitButton`

spli02:


spli02:
Closed this PR as we decided to handle the issue on the solution side instead.

PR #679, opened 2026-06-13
Implement new `Collection` utility and use it to rework `SegmentedControl`

mkrause:
This PR:
- Introduces a new `useFocusGroup` utility which implements a [polyfilled](https://github.com/microsoft/polyfills/tree/main/packages/focusgroup) version of the new [HTML `focusgroup` attribute](https://open-ui.org/components/scoped-focusgroup.explainer).
- Introduces a new `useCollection` hook that can be used for composite components.
- Introduces a new `useSelectionSingle` hook to manage single-selection state.
- Introduces a new `useRadioGroup` hook which builds upon `useCollection` and `useSelectionSingle` for radio group behaviors (including selection management).
- Introduces a new `ControllableState` utility to manage state props that may be either controlled or uncontrolled.
- Adds a new `wrap` prop on `Button` to control whether the contents will wrap on overflow or not (needed for `ToggleButton`).
- Introduces a new `ToggleButton` component to represent a single toggle button which can be toggled on or off.
- Refactors `SegmentedControl` to use `useFocusGroup`, `useRadioGroup`, and `ToggleButton`.

TODO:
- [x] Convert `FocusGroup` component to hook
- [x] Delete old `SegmentedControl` component
- [x] Update `ToggledButton` state prop naming
- [x] Rename `useRadioGroup` to just `useSingleSelect` or something in that sense?
- [x] Deduplicate (non-store related) logic in `Collection` and `RadioGroup`
- [x] Guard against multiple components calling `consumeRegistryChange()` in parallel
- [x] Check `store` prop being passed to DOM in CollectionStore story.
- [x] Issue: `Provider` shorthand in `RadioGroupStore` closes over `context`, so the context won't rerender when the context changes?
- [x] Should nonactive vs. disabled be a thing in SegmentedControl? Maybe it should only ever allow `nonactive` and style it as disabled
- [x] Rethink `useEffectEvent` usage for callback.
- [x] Fix type errors.
- [ ] Check remaining FIXME/TODO comments.
- [ ] Check `SegmentedControl` disabled styling in light mode (color contrast).

Follow ups:
- [x] Update downstream apps `SegmentedControl` usage (legacy prop: `onUpdate`)
- [x] Create follow up ticket for button `wrap` improvements: unconditional rendering of `<span>`, keep the icon aligned with the label even when wrapping (DONE: #700)
- [x] Create ticket to tackle `RadioGroupAsCards`. (DONE: #701)
- [x] Create follow up ticket to implement `useControllableState()` in other components. (DONE: #702)
- [x] Create follow up ticket for updating the controllable state props of other component as per new naming conventions. (DONE: #702)
- [ ] Write tests for React utils
- [ ] Add stories for `useRadioGroup`.


mkrause:
Relevant code pen: https://codepen.io/maikelkrause/pen/gbgweWM

src/components/actions/SegmentedControl/SegmentedControl.tsx
```diff
@@ -0,0 +1,157 @@
...
+      defaultSelected,
+      onSelectedChange,
+      onUpdate, // Legacy alias for `onSelectedChange`
+      disabled = false,
+      nonactive = false,
+      ...propsRest
+    } = props;
+    
+    const segmentedControlContext = useMemoOnce<SegmentedControlContext>(() => ({ size, disabled, nonactive }));
+    const SegmentedControlProvider = useMemoOnce(() => ({ children }: React.PropsWithChildren) =>
```
midhun-pk: Provider is created only once, so it captures the initial context forever. If context changes, the provider still uses the stale one.

src/components/util/collections/CollectionStore.tsx
```diff
@@ -0,0 +1,208 @@
...
+  return context;
+};
+
+export const useCollection = (params: UseCollectionParams = {}) => {
+  const collectionId = React.useId();
+  
+  const store = useMemoOnce(() => createStore(createCollectionSlice({ collectionId })));
+  
+  const context = useMemoOnce(() => ({ store }));
+  const Provider = useMemoOnce(() => ({ children }: React.PropsWithChildren) =>
```
midhun-pk: Same here - Provider is created only once, so it captures the initial context forever. If context changes, the provider still uses the stale one.

src/components/util/collections/RadioGroupStore.tsx
```diff
@@ -0,0 +1,109 @@
...
+  }, [isControlled]);
+  
+  // Note: this context value should be as stable as possible, the state changing means the entire subtree will get
+  // rerendered. The way we've set this up, only a change in `isControlled` will cause this state to change. Changes
+  // to `isControlled` after mount should be avoided by consumers (but are technically allowed).
+  // `props.onStateChange` should be assumed to be unstable. We use `useEffectEvent` to keep a reference to the latest
+  // value of `props.onStateChange` without it being a memo dep. React enforces `useEffectEvent` is not called during
+  // rendering in order to avoid bugs in concurrent rendering. We should call `onStateChange` in event listeners only.
+  const context: RadioGroupContext = React.useMemo(() => ({ store, requestSelect }), [requestSelect]);
+  const Provider = useMemoOnce(() =>
```
midhun-pk: Same here - Provider is created only once, so it captures the initial context forever. If context changes, the provider still uses the stale one.

src/components/util/collections/CollectionStore.tsx
```diff
@@ -0,0 +1,208 @@
...
+export const useCollectionItemWith = <E extends Element>(
+  store: StoreApi<CollectionSlice>,
+  { itemKey }: UseCollectionItemParams,
+) => {
+  const collectionId = useStore(store, state => state.collectionId);
+  const registerItem = useStore(store, state => state.registerItem);
+  const unregisterItem = useStore(store, state => state.unregisterItem);
+  
+  const ref = React.useCallback<React.RefCallback<E>>(el => {
+    if (typeof itemKey === 'undefined') {
```
midhun-pk, line 108: May be this check is not needed as ItemKey is not optional?

src/components/util/collections/CollectionStore.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+type UseCollectionItemResult<E extends Element> = {
+  store: CollectionContext['store'],
+  itemProps: {
+    ref: React.RefCallback<E>,
+    'data-bk-coll-parent': string,
+    'data-bk-coll-item': string,
+  },
+};
+export const useCollectionItem = <E extends Element>(params: UseCollectionItemParams): UseCollectionItemResult<E> => {
```
midhun-pk, line 172: This looks mostly same as the `useCollectionItemWith`. Can we reuse it?

src/components/actions/ToggleButton/ToggleButton.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+type ButtonPropsIrrelevant = 'unstyled' | 'trimmed' | 'kind' | 'variant' | 'asyncTimeout';
+type ToggleButtonProps = Omit<ComponentProps<typeof Button>, ButtonPropsIrrelevant> & {
+  /** Whether this component should be unstyled. */
+  unstyled?: undefined | boolean,
+  
+  // Note: currently only supports `small`, but we may introduce a 'medium' default in the future. Make this field
+  // required for now so that we can change the default later.
+  /** The size of the component. */
+  size: 'small',
```
midhun-pk, line 25: Size should be optional I guess.

src/components/util/collections/CollectionStore.tsx
mkrause: Yes true, but I just added it here as a bit of defensive depth in case the type checker is ignored or otherwise circumvented. I will keep it here, but you're right it's not strictly needed.

src/components/actions/ToggleButton/ToggleButton.tsx
mkrause: I kept it as required, as per the comment above:

```
Make this field required for now so that we can change the default later.
```

I have it as such so that if we introduce a `medium` in the future we can make that the default without it being a breaking change. This is what's in the current `SegmentedControl` also so it's not a change.

src/components/util/collections/CollectionStore.tsx
mkrause: Yes, good catch, I forgot to use `useCollectionItemWith()` here.

src/components/actions/SegmentedControl/SegmentedControl.tsx
```diff
@@ -0,0 +1,154 @@
...
+      selected,
+      defaultSelected,
+      onSelectedChange,
+      onUpdate, // Legacy alias for `onSelectedChange`
+      disabled = false,
+      nonactive = false,
+      ...propsRest
+    } = props;
+    
+    const segmentedControlContext = useMemoOnce<SegmentedControlContext>(() => ({ size, disabled, nonactive }));
```
midhun-pk: Q. This should not be memoized only once i guess?

midhun-pk (approved):

PR #683, opened 2026-07-01
Update page index indicator when changing page size

### LINKED ISSUE #545: DataTable: changing page size doesn't properly update the current page number
To reproduce:

1. Take a `DataTableEager` with a lot of pages, for example [this story](https://fortanix.github.io/baklava/?path=/docs/components-tables-datatable-datatableeager--docs#multiple-pages-large).
2. Go to a subsequent page (e.g. page 3)
3. Increase the page size (e.g. 50 instead of 10)
4. Notice that the page size has increased, but the page number still reflects the old state (e.g. if it was page 3 before it's still on page 3 after)

nighto:
Fixes #545

Steps to reproduce:
1. Open multiple pages large example, change page to 50
image

2. Switch rows per page to another value, for instance 100, page gets updated:
image

### DIFF
--- a/src/components/tables/DataTable/pagination/PaginationSizeSelector.tsx
+++ b/src/components/tables/DataTable/pagination/PaginationSizeSelector.tsx
@@ -19,6 +19,7 @@ export const defaultPageSizeOptions: Array<PageSizeOption> = [10, 25, 50, 100];
 type PaginationSizeSelectorProps = {
   pageSizeOptions?: undefined | Array<PageSizeOption>,
   pageSizeLabel?: undefined | string,
+  setPageIndexIndicator?: undefined | React.Dispatch<React.SetStateAction<number>>,
 };
 export const PaginationSizeSelector = (props: PaginationSizeSelectorProps) => {
   const { pageSizeOptions = defaultPageSizeOptions, pageSizeLabel = 'Rows per page' } = props;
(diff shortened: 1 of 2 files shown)


src/components/tables/DataTable/pagination/PaginationSizeSelector.tsx
```diff
@@ -39,6 +40,15 @@ export const PaginationSizeSelector = (props: PaginationSizeSelectorProps) => {
...
               table.setPageSize(pageSize);
+              if (props.setPageIndexIndicator) {
+                // manually update the page index
+                const oldPageSize = table.state.pageSize;
+                const newPageSize = pageSize;
+                const oldPageIndex = table.state.pageIndex;
+                const newPageIndex = Math.floor(oldPageIndex * oldPageSize / newPageSize);
+                // page displayed is always +1 than the index
+                props.setPageIndexIndicator(newPageIndex + 1);
+              }
```
mkrause: I think this needs to be fixed elsewhere. The table state is being updated already as part of `table.setPageSize(pageSize);`. So the question is, why isn't that table state being synchronized properly?

mkrause (changes requested):

src/components/tables/DataTable/pagination/PaginationSizeSelector.tsx
nighto: I am trying to get my head around this but I can't. If we manually call `setPageIndexIndicator` on the manual page changes, be it the first / previous / next / last buttons (`Pagination.tsx` lines 60, 71, 100 and 111) as well as on the manual page change by changing it's number (line 84), why shouldn't we do the same here?

mkrause: @nighto the issue is that we have two pieces of state, one is `table.state` and the other is this local `pageIndexIndicator` state. I'm not sure why we have the `pageIndexIndicator` state in the first place, if we can remove that then the state can never get out of sync in the first place (we'd have just one source of truth).

```ts
  const { table } = useTable();
  const [pageIndexIndicator, setPageIndexIndicator] = React.useState<number>(1);
  /*
  Available pagination state:
  - table.state.pageIndex
  - table.state.pageSize
  - table.canPreviousPage
  - table.canNextPage
  - table.pageOptions
  - table.pageCount
  - table.gotoPage
  - table.nextPage
  - table.previousPage
  - table.setPageSize
  */
  ```

I would suggest:

- Remove the `useState` here.
- `pageIndexIndicator` would just be `table.state.pageIndex + 1`.
- Refactor the `updatePage` logic so that instead of deriving it from `pageIndexIndicator` state, it just works with the `table` API directly.

mkrause (approved):

PR #686, opened 2026-07-04
Refactor `ListBox` and related components – New `MenuList` component, on top of which `Menu` and `ListBox`

mkrause:
Breaking changes:
- `ListBox` hidden form input is now not present when no item is selected
- `ListBox` `onSelect` is deprecated, use `onSelectedChange` instead
- `ListBox.Option` `onSelect` prop was removed, use the parent `ListBox` `onSelectedChange` callback instead
- `ListBox.Header` is replaced with `ListBox.Group`
- `ListBox.Action`, `ListBox.FooterAction`, and `ListBox.FooterActions` have been removed. The new `Menu` component should be used for this instead.

Components:
- [x] MenuList
    - [x] Add an `embedded` variant
    - [x] Move `iconDecoration` highlight to `Button`
    - [x] Fix `MenuList` "footer"
    - [x] `MenuList`: icons
    - [ ] `MenuList`: "chin" should remain visible while scrolling with a `sticky="end"` footer
    - [ ] `MenuList` idea: automatically calculate the sticky header/footer size so we don't need to manually specify `--sticky-items-start` and `--sticky-items-end`?
    - [ ] `Group` heading: fix all-caps accessible name (introduce `Text` component?)
- [ ] ListBox + ListBoxMulti
    - [x] Re-implement type-ahead search + focusFirst/focusLast.
    - [ ] Figure out if we still need `formatItemKey`.
    - [x] Reconsider if we should use `React.memo()`. Better to guide consumers to memo `children`?
- [ ] ListBoxLazy + ListBoxMultiLazy
- [ ] Menu
    - [x] Add a `Menu` element that can wrap an embedded `ListBox` and add actions, search inputs, etc.
    - [ ] How should the `empty` logic work for `Menu`? An `Action` and `Group` don't count as a collection item currently. But `Menu` doesn't support `Option` at all.
- [ ] MenuProvider
- [ ] Reinstate `onActive` instead of `onPress`?

To do:
- [ ] Open tasks from https://github.com/fortanix/baklava/pull/679
- [x] Change `isLoading` to a "status" prop enum?
- [x] `Button` should have a "basic" variant that is unstyled but still supports things like `wrap`?

Follow ups:
- [ ] Update downstream applications for any breaking changes.
- [ ] Remove all unnecessary `null | ` in `xgrep 'useRef<null'`
- [ ] Change `orientation` props to logical variants?
- [ ] Do rerender testing, add stories for memoing `children` in case of large amounts of items.

BREAKING:
- Custom icon prop `Icon` changed.
- `isLoading` changed to `status`.
- Removed deprecated `action` props on `MenuProvider` (and variants).


src/components/util/collections/ListBoxMultiStore.ts
```diff
@@ -0,0 +1,147 @@
...
+
+
+type UseListBoxMultiItemParams = { itemKey: ItemKey };
+export const useListBoxMultiItem = <E extends HTMLElement = HTMLElement>({ itemKey }: UseListBoxMultiItemParams) => {
+  const { store, requestSelected } = useListBoxMultiContext();
+  const selected = useStore(store, store => store.selectedItemKeys?.has(itemKey) ?? false);
+  
+  // FIXME: better way to add/remove item keys from the store. Should `requestSelected` take a callback?
+  // FIXME: "requestSelected" doesn't make sense for multi where things can be unselected as well as selected
+  const selectedItemKeys = useStore(store, store => store.selectedItemKeys);
```
midhun-pk, line 132: Do we have to subscribe to selectedItemKeys in the store here? It will basically re-render every item in the list i guess? Can we just read the value instead?

Like `store.getState().selectedItemKeys()`

src/components/util/collections/CollectionStore.ts
```diff
@@ -0,0 +1,357 @@
...
+    const query: string = removeCombiningCharacters(sequence.join(''));
+    if (query.trim() === '') { return; }
+    
+    let itemEls = getItemElements();
+    const focusedElement = document.activeElement instanceof HTMLElement ? document.activeElement : null;
+    
+    // Cycle the elements such that the focused item (if any) comes first. This is so that the type-ahead search will
+    // always continue from the current focused element.
+    const focusedNodeIndex = focusedElement ? itemEls.indexOf(focusedElement) : -1;
+    if (focusedNodeIndex) {
```
midhun-pk, line 322: Q. The condition should be `focusedNodeIndex > 0` i guess?

midhun-pk (changes requested):

PR #692, opened 2026-07-10
Release/v1.0.0 beta 20260710

nighto:


CONTRIBUTING.md
```diff
@@ -19,7 +19,7 @@ format `<username>/<date>-<summary>`, for example `mkrause/241129-upgrade-depend
 To create a new release:
 
 - First, create a release PR:
-  - Create a release branch named `release/vx.y.z`.
+  - Create a release branch named `release/v1.0.0-beta-yyyymmdd`.
```
mkrause: I don't think we need to specify it this specifically. But also there's other references to "x.y.z" below so now it's inconsistent. I think we can just leave it as x.y.z

mkrause (approved):

PR #697, opened 2026-07-16
Stepper component improvements

### LINKED ISSUE #262: `Stepper` improvements
- [ ] Use composition instead of the `steps` prop (e.g. a `Stepper.Step` subcomponent)
- [ ] Responsive design
- [ ] Fix issue where if a step label wraps over multiple lines, the connecting line between steps becomes broken

Mahek-Vasoya:
Implemented stepper component with anchor positioning.

Resolves #262

### DIFF
--- a/src/components/navigations/Stepper/Stepper.module.scss
+++ b/src/components/navigations/Stepper/Stepper.module.scss
@@ -144,6 +144,11 @@
         align-self: center;
         white-space: nowrap;
       }
+      
+      .bk-stepper__step__optional {
+        margin-inline-start: bk.$spacing-2;
+        font-size: bk.$font-size-xs;
+      }
 
       .bk-stepper__step__body {
         anchor-name: --step-body;
--- a/src/components/navigations/Stepper/Stepper.module.scss
+++ b/src/components/navigations/Stepper/Stepper.module.scss
@@ -11,7 +11,7 @@
     --bk-stepper-indicator-size: #{bk.rem-from-px(28)};
     --bk-stepper-line-width: #{bk.$size-1};
     
-    >ol {
+    > ol {
       display: flex;
       flex-direction: column;
       gap: bk.$spacing-9;
@@ -20,9 +20,11 @@
     }
     
     &.bk-stepper--horizontal {
+      padding-block: bk.$spacing-1;
+      padding-inline-start: bk.$spacing-1;
       overflow-y: hidden;
       
-      >ol {
+      > ol {
         flex-direction: row;
         flex-wrap: nowrap;
         min-inline-size: 100%;
@@ -32,26 +34,23 @@
         position: relative;
         flex: 0 0 auto;
 
-        &:last-child {
-          flex: 1 0 auto; // Grow only the last step.
-        }
-
         &:not(:last-child)::after {
           content: '';
 
           position: absolute;
 
-          inset-inline-start: anchor(--step-action right);
+          inset-inline-start: calc(anchor(--step-action right) + bk.$spacing-2);
           inset-block-start: calc(anchor(--step-action top) + var(--bk-stepper-indicator-size) / 2);
 
-          inline-size: calc(anchor-size(--step-action inline) - anchor-size(--step-label inline));
+          inline-size: calc(anchor-size(--step-action inline) - anchor-size(--step-label inline) - bk.$spacing-2);
           block-size: var(--bk-stepper-line-width);
 
           background: bk.$theme-stepper-border-disabled;
         }
 
         &:has(.bk-stepper__step__body)::after {
-          inline-size: calc(anchor-size(--step-body inline) - anchor-size(--step-label inline) + bk.$spacing-9);
+          inline-size: calc(anchor-size(--step-body inline) - anchor-size(--step-label inline) + bk.$spacing-9 
+          - bk.$spacing-2);
         }
 
         .bk-stepper__step__action {
@@ -129,10 +128,6 @@
         border-radius: 50%;
         font-weight: bk.$font-weight-bold;
         font-size: bk.$font-size-m;
-
-        &:empty::before {
-          content: counter(list-item);
-        }
 
         .bk-stepper__step__indicator__icon {
           font-size: bk.$font-size-xs;
--- a/src/components/navigations/Stepper/Stepper.module.scss
+++ b/src/components/navigations/Stepper/Stepper.module.scss
@@ -9,97 +9,173 @@
     @include bk.component-base(bk-stepper);
     
     --bk-stepper-indicator-size: #{bk.rem-from-px(28)};
-    display: flex;
+    --bk-stepper-line-width: #{bk.$size-1};
     
-    > ol {
-      display: contents;
+    >ol {
+      display: flex;
+      flex-direction: column;
+      gap: bk.$spacing-9;
+
+      anchor-scope: --step-link;
     }
     
     &.bk-stepper--horizontal {
-      flex-direction: row;
-      column-gap: bk.$spacing-9;
-    }
-    
-    &.bk-stepper--vertical {
-      flex-direction: column;
-      row-gap: bk.$spacing-9;
+      overflow-y: hidden;
       
-      // Draw a line between subsequent items
-      li + li .bk-stepper__item__indicator {
-        &::before {
+      >ol {
+        flex-direction: row;
+        flex-wrap: nowrap;
+        min-inline-size: 100%;
+      }
+
+      .bk-stepper__step {
+        position: relative;
+        flex: 0 0 auto;
+
+        &:last-child {
+          flex: 1 0 auto; // Grow only the last step.
+        }
+
+        &:not(:last-child)::after {
           content: '';
+
           position: absolute;
-          inset-block-start: calc(-1 * bk.$spacing-9 - bk.$size-2);
-          inset-inline-start: calc(50% - bk.$size-2 / 2);
-          inline-size: 0;
-          block-size: bk.$spacing-9;
-          border-inline-start: bk.$size-2 solid bk.$theme-stepper-border-disabled;
+
+          inset-inline-start: anchor(--step-action right);
+          inset-block-start: calc(anchor(--step-action top) + var(--bk-stepper-indicator-size) / 2);
+
+          inline-size: calc(anchor-size(--step-action inline) - anchor-size(--step-label inline));
+          block-size: var(--bk-stepper-line-width);
+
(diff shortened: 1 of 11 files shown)


mkrause:
The focus outline is cut off:

Screenshot 2026-07-17 at 01 40 08

mkrause:
For the horizontal version, can we add a bit of spacing between the content and the line? Currently they're flush against each other.

src/components/navigations/Stepper/Stepper.module.scss
```diff
@@ -9,97 +9,173 @@
     @include bk.component-base(bk-stepper);
     
     --bk-stepper-indicator-size: #{bk.rem-from-px(28)};
-    display: flex;
+    --bk-stepper-line-width: #{bk.$size-1};
     
-    > ol {
-      display: contents;
+    >ol {
```
mkrause: Nit: space between the `>` and `ol`

src/components/navigations/Stepper/Stepper.module.scss
```diff
@@ -9,97 +9,173 @@
...
+        flex-wrap: nowrap;
+        min-inline-size: 100%;
+      }
+
+      .bk-stepper__step {
+        position: relative;
+        flex: 0 0 auto;
+
+        &:last-child {
+          flex: 1 0 auto; // Grow only the last step.
```
mkrause: Is this needed? What happens if the last item doesn't expand?

src/components/navigations/Stepper/Stepper.module.scss
```diff
@@ -9,97 +9,173 @@
...
         font-weight: bk.$font-weight-bold;
         font-size: bk.$font-size-m;
-        
-        display: grid;
-        place-content: center;
-        
-        .bk-stepper__item__indicator__icon {
+
+        &:empty::before {
+          content: counter(list-item);
```
mkrause, line 134: Are we actually using this? I thought we're always setting the count explicitly in JS.

mkrause (changes requested):

src/components/navigations/Stepper/Stepper.module.scss
Mahek-Vasoya: yes this styles were left behind while cleaning up. Thanks for highlighting

src/components/navigations/Stepper/Stepper.module.scss
Mahek-Vasoya: same here

src/components/navigations/Stepper/Stepper.tsx
```diff
@@ -3,104 +3,299 @@
...
+          {...(orientation === 'horizontal' ? scrollerProps : {})}
+          className={cx(
+            'bk',
+            { [cl['bk-stepper']]: !unstyled },
+            { [cl['bk-stepper--vertical']]: orientation === 'vertical' },
+            { [cl['bk-stepper--horizontal']]: orientation === 'horizontal' },
+            propsRest.className,
+          )}
+        >
+          <div className={cx("connector-line")}></div>
```
mkrause: Is this intentional? This doesn't use `cl[]`, and it's not referenced anywhere.

mkrause (changes requested):

src/components/navigations/Stepper/Stepper.tsx
```diff
@@ -3,104 +3,299 @@
...
+    }
+
+    if (child.type === React.Fragment) {
+      const [fragmentChildren, updatedCount] = assignCounts(child.props.children, nextCount, reverse);
+      nextCount = updatedCount;
+
+      return (
+        <React.Fragment>
+          {fragmentChildren}
+        </React.Fragment>
```
mkrause: Can just write it as `<></>`

src/components/navigations/Stepper/Stepper.tsx
```diff
@@ -3,104 +3,262 @@
...
+- `aria-current="step"` should be applied to the `<li>` that is the currently active step.
+- Some `class="visually-hidden"` text elements should be added to each step to clarify the state (e.g. "Completed").
 */
 
-export { cl as SteppersClassNames };
+export { cl as StepperClassNames };
+
+
+
+type StepKey = string;
```
mkrause: Let's export this type.

src/components/navigations/Stepper/Stepper.tsx
```diff
@@ -3,104 +3,262 @@
...
+
+//
+// Context
+//
+
+export type StepperContext = {
+  activeStepKey: StepKey | undefined;
+  setActiveStepKey: (stepKey: StepKey) => void;
+  completedStepKeys: ReadonlySet<StepKey>;
+};
```
mkrause, line 47: Nit: `,` instead of `;` for delimiters.

src/components/navigations/Stepper/Stepper.tsx
```diff
@@ -3,104 +3,262 @@
...
+  stepKey: StepKey,
+
+  /** The human-readable name for this step. */
+  label: string;
+
+  /** Additional content rendered below the label. */
+  children?: React.ReactNode;
+
+  /** Override the displayed step number. */
+  count?: undefined | number;
```
mkrause: In the above few props also: `,` instead of `;`

src/components/navigations/Stepper/Stepper.tsx
```diff
@@ -3,104 +3,262 @@
...
-        { [cl['bk-stepper']]: !unstyled },
-        { [cl['bk-stepper--horizontal']]: direction === 'horizontal' },
-        { [cl['bk-stepper--vertical']]: direction === 'vertical' },
+        { [cl['bk-stepper__step']]: !unstyled },
+        { [cl['bk-stepper__step--completed']]: isCompleted },
+        { [cl['bk-stepper__step--disabled']]: disabled },
         propsRest.className,
       )}
+      style={{
+        counterSet: count != null ? `list-item ${count}` : undefined,
```
mkrause: Can we change this check to: `typeof count === 'number'` ?

src/components/navigations/Stepper/Stepper.tsx
```diff
@@ -3,104 +3,262 @@
...
+  unstyled?: undefined | boolean,
+
+  /** A unique human-readable name for this landmark. Required. */
+  label: string,
+
+  /** Whether this component should be displayed vertically or horizontally. Default: `"vertical"`. */
+  orientation?: undefined | 'vertical' | 'horizontal',
+
+  /** Controlled active step. */
+  activeStepKey?: undefined | StepKey;
```
mkrause: In these also: `,` instead of `;`

src/components/navigations/Stepper/Stepper.tsx
```diff
@@ -3,104 +3,262 @@
...
+    } = props;
+
+    const scrollerProps = useScroller();
+    const isControlled = controlledActiveStepKey !== undefined;
+    const [uncontrolledActiveStepKey, setUncontrolledActiveStepKey] = React.useState(defaultActiveStepKey);
+
+    const activeStepKey = isControlled ? controlledActiveStepKey : uncontrolledActiveStepKey;
+
+    const stepKeys = React.useMemo(() => {
+      const keys: StepKey[] = [];
```
mkrause: Nit: `Array<StepKey>`

src/components/navigations/Stepper/Stepper.tsx
```diff
@@ -3,104 +3,262 @@
...
+
+      React.Children.forEach(children, child => {
```
mkrause: This is still using `React.Children`. We should use `CollectionStore` instead.

src/components/navigations/Stepper/Stepper.tsx
```diff
@@ -3,104 +3,262 @@
...
+  defaultActiveStepKey?: undefined | StepKey,
+
+  /** Callback executed when active step is changed. */
+  onSwitch?: undefined | ((stepKey: StepKey) => void),
+
+  /** The starting number of the list (if different from 1). Optional. */
+  start?: undefined | number,
+
+  /** Whether the ordered list should be rendered in reverse order. */
+  reverse?: undefined | boolean;
```
mkrause: Should we make this `reversed` to make it consistent with the native DOM one?

mkrause (changes requested):

src/components/navigations/Stepper/Stepper.tsx
```diff
@@ -3,104 +3,262 @@
...
-        'bk',
-        { [cl['bk-stepper']]: !unstyled },
-        { [cl['bk-stepper--horizontal']]: direction === 'horizontal' },
-        { [cl['bk-stepper--vertical']]: direction === 'vertical' },
+        { [cl['bk-stepper__step']]: !unstyled },
+        { [cl['bk-stepper__step--completed']]: isCompleted },
+        { [cl['bk-stepper__step--disabled']]: disabled },
         propsRest.className,
       )}
+      style={{
```
mkrause, line 128: Add a comment here explaining why we do, that it affects Chrome v150 but not v152, and to remove this once browser support is good enough.

mkrause (changes requested):

src/components/navigations/Stepper/Stepper.tsx
```diff
@@ -3,104 +3,258 @@
...
-  
+
+  /** The key for this step, should be unique within the `Stepper` component. */
+  stepKey: StepKey,
+
+  /** The human-readable name for this step. */
+  label: string,
+
+  /** Additional content rendered below the label. */
+  children?: React.ReactNode,
```
mkrause, line 75: Here, the `children` are used as essentially the "description" of the step. But maybe it's better to have a separate `description?: undefined | React.ReactNode` prop for this? Which would free up `children` to be an optional `ReactNode` version of `label` in the same way that it's done for `Button` and some other components.

Currently, a `Step` cannot have a `label` which is anything other than a `string`.

mkrause (approved):

mkrause (approved):

src/components/navigations/Stepper/Stepper.tsx
```diff
@@ -3,104 +3,269 @@
...
-  onSwitch: (stepKey: StepperKey) => void,
-}>;
-/**
- * A stepper component
- */
-export const Stepper = (props: StepperProps) => {
-  const { unstyled = false, steps = [], activeKey, direction = 'vertical', onSwitch, ...propsRest } = props;
-  
+  /** Whether this step should optional. Default: `false`. */
+  isOptional?: undefined | boolean,
```
mkrause: Can you rename to `optional`?

PR #698, opened 2026-07-20
NPM security hardening

Mahek-Vasoya:
A transitive dependency currently resolves to less@4.5.1, which includes an unintended postinstall script that attempts to install Playwright. This was acknowledged by the Less maintainers as a packaging bug and removed in less@4.6.0+.

With npm's install script allowlisting enabled, this unnecessary postinstall causes installation to fail unless explicitly approved.

Using an npm overrides entry ensures that all transitive dependencies resolve to less@4.6.7, avoiding the unwanted postinstall while preserving the existing dependency graph.

**Impact**
Removes the unnecessary postinstall script from the dependency tree.
Eliminates the need to approve less via allowScripts.
Uses the latest patched Less 4.x release without requiring updates to upstream dependencies.

**References :** Less issue: https://github.com/less/less.js/issues/4396


imadbakir (approved):

.github/workflows/ci.yaml
```diff
@@ -27,6 +27,8 @@ jobs:
           node-version: ${{ matrix.node-version }}
           cache: 'npm'
       - run: npm ci
+      - run: node -v
+      - run: npm -v
```
mkrause: Should we have these here? Not sure if this is just for debugging or not.

mkrause (changes requested):

.github/workflows/ci.yaml
Mahek-Vasoya: Yes, this was added intentionally earlier when we had the range for this, but now we can remove it.

mkrause (approved):

Issue #728, opened 2026-09-16
Upgrade typescript to v7


nighto:

nighto:
As an intermediate step, upgrade to v6, since some breaking changes on v7 are considered deprecated on v6.

PR #729, opened 2026-09-16
Upgrade typescript to v6

nighto:
This change was done with GitHub Copilot, interacting with the CLI prompt (I was curious to try it out). The tool described the changes done (outside of the trivial changes), including changing some deprecated ones.

storybook:serve runs fine, and npm run test also runs fine, with no difference on the tests passed (all pass) and the time to run (virtually the same).

We might consider whether it is better to keep both `.github/copilot-instructions.md` and `.github/workflows` on versioning or add them to `.gitignore`.

These specific changes, this was the CLI output for each of them:

- Removed redundant deprecated `esModuleInterop: false` from `tsconfig.app.json` and `tsconfig.node.json`.

> TypeScript configs to remove the deprecated  esModuleInterop: false

- Added the `virtual:svg-icons/register` ambient declaration in `tests/installation/src/vite-env.d.ts`.

> TypeScript 6 is installed and the installation lockfile is refreshed. The consumer smoke test now fails on a stricter side-effect-import check for the existing  virtual:svg-icons/register  import, so I’m checking the consumer project’s declarations before deciding whether a small compatibility fix is needed.

- Updated `src/components/util/overlays/popover/usePopover.ts` to omit `togglePopover`'s `source` property when undefined, preserving behavior under `exactOptionalPropertyTypes`.

> TypeScript 6’s DOM types now reject passing  source: undefined  to  togglePopover  under this repository’s  exactOptionalPropertyTypes  setting. The fix preserves behavior by omitting the  source  property when no valid source element exists; I’ll apply that compatibility change and rerun validation.

Issue #728


.github/checkpoints/2026-09-16-typescript-6-upgrade.md
```diff
@@ -0,0 +1,41 @@
+# Checkpoint: TypeScript 6 Upgrade
```
mkrause: Let's not commit these two `.md` files

src/components/util/overlays/popover/usePopover.ts
```diff
@@ -51,14 +51,14 @@ const syncPopoverWithController = (params: {
   const isPopoverActive = popoverEl.matches(':popover-open');
   if (controllerActive && !isPopoverActive) { // Case: should be active but isn't
     try {
-      popoverEl.togglePopover({ force: true, source });
+      popoverEl.togglePopover(source === undefined ? { force: true } : { force: true, source });
```
mkrause: I'm not sure why these changes are needed. The type is defined here in `globals.d.ts`:

```
type TogglePopoverOptions = {
  force?: undefined | boolean,
  // NOTE: `source` must be an `HTMLElement`, not just any `Element`. If given, for example, an `<svg>` element, the
  // browser will throw an exception.
  source?: undefined | HTMLElement,
};
```

So `source` can already be unspecified (`?:`) or explicitly `undefined`. So the existing code should work as far as I can tell.

EDIT: it might be that due to TS now supporting `source` out of the box in its `lib.d.ts`, and that this is conflicting. In which case we should re-evaluate the types in `globals.d.ts` and see if we need to remove (some of) these.

mkrause (changes requested):

src/components/util/overlays/popover/usePopover.ts
nighto: Yeah that seems to be the case. When undoing this specific change, VSCode errors out:

Image

Just to double check, it is indeed using `tsc` as defined on the repo:

Image

Full error text (as it is cut out on image above):

```
No overload matches this call.
  Overload 1 of 2, '(options?: TogglePopoverOptions | undefined): boolean', gave the following error.
    Argument of type '{ force: true; source: HTMLElement | undefined; }' is not assignable to parameter of type 'TogglePopoverOptions' with 'exactOptionalPropertyTypes: true'. Consider adding 'undefined' to the types of the target's properties.
      Types of property 'source' are incompatible.
        Type 'HTMLElement | undefined' is not assignable to type 'HTMLElement'.
          Type 'undefined' is not assignable to type 'HTMLElement'.
  Overload 2 of 2, '(options?: boolean | TogglePopoverOptions | undefined): boolean', gave the following error.
    Argument of type '{ force: true; source: HTMLElement | undefined; }' is not assignable to parameter of type 'boolean | TogglePopoverOptions | undefined'.
      Type '{ force: true; source: HTMLElement | undefined; }' is not assignable to type 'TogglePopoverOptions' with 'exactOptionalPropertyTypes: true'. Consider adding 'undefined' to the types of the target's properties.
        Types of property 'source' are incompatible.
          Type 'HTMLElement | undefined' is not assignable to type 'HTMLElement'.
            Type 'undefined' is not assignable to type 'HTMLElement'.
```

mkrause (approved):

